PR #38, opened 2015-08-30
Support setting all actions (not only modify, create and delete).

### LINKED ISSUE #37: Ac Tool only set modify, create, delete actions in repository
Since 1.6.1 release, the AC Tool only sets modify, create, delete cq-actions in repository. Other cq-actions like read doesn't get set anymore.

kwin:
In addition support setting restrictions on actions (which is not
supported by the CQ Security API)
Fixes #37

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/AceBean.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/AceBean.java
@@ -322,12 +322,12 @@ public class AceBean implements AcDumpElement {
 		cqActions.installActions(getJcrPath(), principal, actionMap,
 				inheritedAllows);
 		
-		Map<String, Value> restrictions = getSingleValueRestrictions(session, acl);
-		if (restrictions.isEmpty()) {
-			return acl;
-		}
 		// since the aclist has been modified, retrieve it again
 		JackrabbitAccessControlList newAcl = AccessControlUtils.getAccessControlList(session, getJcrPath());
+		Map<String, Value> restrictions = getSingleValueRestrictions(session, acl);
+		if (restrictions.isEmpty()) {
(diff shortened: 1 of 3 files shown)


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/AceBean.java
```diff
@@ -249,47 +259,101 @@ private void removeRedundantPrivileges(Session session) throws RepositoryExcepti
...
-						break;
-					}
-				}
+		CqActions cqActions = new CqActions(session);
+		Collection<String> inheritedAllows = cqActions.getAllowedActions(
+				getJcrPath(), Collections.singleton(principal));
+		// this does always install new entries
+		cqActions.installActions(getJcrPath(), principal, actionMap,
+				inheritedAllows);
+
```
jochenkoschorke: Since there are potential changes in the acl done by the call to cqActions.installActions(), the acl has to be retrieved again directly afterwards. Otherwise the old ace gets returned in case only actions (without globbing, or additional privileges) would have been set. So, the later call to: AccessControlUtils.getAccessControlList(session, getJcrPath()) has to be moved up.

kwin: This would not matter for getSingleValueRestrictions because the given acl is only used to check, which restriction names are supported at all. Not which ones are already set.

jochenkoschorke: Yes, it would not matter for that purpose, but the old acl would get returned by AceBean. installActions(). The subsequent call to acMgr.setPolicy(getJcrPath(), acl); (in line 401, method AceBean.install()) would use the old acl in the described case.

Issue #54, opened 2015-12-03
Deploy to CQ5.6.1 fails due to AEM6 dependencies and Oak namespace


cwoeltge:
Deploying the package to CQ5.6.1 fails with a namespace exception due Oak index definition package. Got that sorted but bundle dependencies point to AEM6 and the bundle won't start.

kwin:
Closed with merging PR #55.

kwin:
Actually there is still one incompatibility in the code:
In https://github.com/Netcentric/accesscontroltool/blob/develop/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/AceWrapper.java#L127 we implement a `JackrabbitAccessControlEntry.getRestrictions(String)` method which was only implemented in jackrabbit-api 2.7.1 (https://issues.apache.org/jira/browse/JCR-3637). CQ 5.6.1 ships though with jackrabbit-api 2.6.2.

During bundle start there is no problem though, because for the change in JCR-3637 the package version has no been incremented correctly (it is version 2.3.0 with and without that newly added method). That is a violation of semantic versioning, therefore you can see the problem only during run time.

mtstv:
Is it not the time now to start to split the code to different releases for 5.6.1, 6.0 (SP2, SP3), 6.1?

kwin:
I want to prevent maintaining different branches of the code. Rather we should focus on providing compatibility with a reasonably old CQ version (which in this case is CQ 5.6.1). Maybe there is another solution which is compatible with AEM 6.1 as well as CQ 5.6.1 (maybe by getting rid of the ACEWrapper or by let that just embed the original JackrabbitAccessControlEntry).

cwoeltge:
Actually, it should work on 5.6.1 - with proper Service Pack installed Jackrabbit API version is 2.7.5 - let's make SP2 the restriction (https://docs.adobe.com/content/docs/en/cq/5-6-1/release_notes_service_pack_2.html)

kwin:
I agree.
I will update the readme and the pom.xml accordingly then.

mtstv:
I am not sure, but may be you can try to compile with uber-jar artefact to reduce the set of dependencies?
https://repo.adobe.com/nexus/content/repositories/releases/com/adobe/aem/uber-jar/

PR #55, opened 2015-12-03
#54: make CQ5.6.1 the default dependency

cwoeltge:
I've removed all runtime dependencies to 6.x to ensure the bundle starts under CQ5.6.1

Oak Index package has been moved to a separate profile `oakindex`

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,8 +25,11 @@ The package can be installed using the AEM Package Manager or directly from the 
 mvn -PautoInstallPackage install
 ```
 
-## AEM6.x
-To add an Oak Index install, run all commands with profile `oakindex`, e.g.
+## AEM6.x/Oak
+
+The `oakindex-package` contains an optimized Oak index to cover all queries being issued by the Access Control Tool. To build (and optionally deploy) the content-package use the Maven profile oakindex. This package is only compatible with Oak and even there it is optional (as it will only speed up queries).
+
+To use the package, run all commands with profile `oakindex`, e.g.
  ```
 mvn clean install -Poakindex
  ```
--- a/README.md
+++ b/README.md
@@ -25,8 +25,11 @@ The package can be installed using the AEM Package Manager or directly from the 
 mvn -PautoInstallPackage install
 ```
 
-## Working on CQ5.6
-To deploy to CQ5.6 switch to branch `develop-cq56`.
+## AEM6.x
+To add an Oak Index install, run all commands with profile `oakindex`, e.g.
+ ```
+mvn clean install -Poakindex
+ ```
 
 # Configuration File Format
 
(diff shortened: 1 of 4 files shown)


accesscontroltool-bundle/pom.xml
```diff
@@ -121,6 +117,24 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.apache.sling</groupId>
+            <artifactId>org.apache.sling.settings</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.day.jcr.vault</groupId>
+            <artifactId>com.day.jcr.vault</artifactId>
+        </dependency>
+        <dependency>
```
kwin: Please rather get rid of this dependency (even though currently only constants are referenced which are inlined anyways). It is fine to use those constants in the test classes (so leave it with scope test) but in the main classes we should not put this dependency (otherwise too easily a real runtime dependency towards Oak may slip in the code)

pom.xml
```diff
@@ -76,17 +76,33 @@
...
+                <artifactId>com.day.jcr.vault</artifactId>
+                <version>2.4.32</version>
+            </dependency>
+            <dependency>
+                <groupId>com.day.cq</groupId>
+                <artifactId>cq-commons</artifactId>
+                <version>5.6.4</version>
                 <scope>provided</scope>
             </dependency>
+            <!-- 5.6.1 dependencies end -->
```
kwin: for me even the following dependencies must have the 5.6.1 version and therefore are 5.6.1 dependencies!

cwoeltge: It was the area replacing the aem 6.0.0 dependency - but I agree it doesn't seem to make a lot of sense now. Let's remove the comment at all.

README.md
```diff
@@ -25,6 +25,12 @@ The package can be installed using the AEM Package Manager or directly from the
 mvn -PautoInstallPackage install
 ```
 
+## AEM6.x
```
kwin: What about making this headline say "Oak" and some more information about this Oak Index extension like. "The oakindex-package contains an optimized Oak index to cover all queries being issued by the Access Control Tool. To build (and optionally deploy) the content-package use the Maven profile `oakindex`. This package is only compatible with Oak and even there it is optional (as it will only speed up queries).

pom.xml
```diff
@@ -174,7 +188,7 @@
             <dependency>
                 <groupId>org.apache.jackrabbit</groupId>
                 <artifactId>jackrabbit-api</artifactId>
-                <version>2.3.0</version>
+                <version>2.7.5</version>
```
kwin: Only version 2.6.2 is contained in 5.6.1

PR #63, opened 2016-03-09
Upgrade Apache Commons Collections to v3.2.2

gmlewis:
Version 3.2.1 has a CVSS 10.0 vulnerability. That is the worst kind of
vulnerability that exists. By merely existing on the classpath, this
library causes the Java serialization parser for the entire JVM process
to go from being a state machine to a turing machine. A turing machine
with an exec() function!

https://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2015-8103
https://commons.apache.org/proper/commons-collections/security-reports.html
http://foxglovesecurity.com/2015/11/06/what-do-weblogic-websphere-jboss-jenkins-opennms-and-your-application-have-in-common-this-vulnerability/

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -180,7 +180,7 @@
             <dependency>
                 <groupId>commons-collections</groupId>
                 <artifactId>commons-collections</artifactId>
-                <version>3.2.1</version>
+                <version>3.2.2</version>
                 <scope>provided</scope>
             </dependency>
             <dependency>


kwin:
Actually actool is not embedding commons collections in any way. It only depends on it! And depending on 3.2.1 it totally fine here. Basically only AEM (as the one providing that dependency) needs to be fixed. This is done with hotfix 8364 for AEM 6.0 and SP1 for AEM 6.1 (more information available at https://helpx.adobe.com/experience-manager/kb/aem6-available-hotfixes.html and https://helpx.adobe.com/experience-manager/kb/aem61-available-hotfixes.html).

davidjgonzalez:
FYI - i think this is a bot that is crawling GH

gmlewis:
Sorry, no, it is not a bot, but I am processing many repos.  I apologies for the noise.

Issue #69, opened 2016-04-15
Allow moving of users in the same way as it is possible for groups


ghenzler:
... and get rid of the log entry "authorizable xxx exists and is a user - moving and updating of users is not implemented yet.

PR #70, opened 2016-04-15
#69 support moving of users

ghenzler:


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableutils/AuthorizableConfigBean.java
```diff
@@ -28,7 +28,6 @@
     private String[] members;
     String membersStringFromConfig;
 
-    private String description;
```
ghenzler: This was not used anywhere - we should introduce a more generic approach for setting profile information (created #71)

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableutils/impl/AuthorizableCreatorServiceImpl.java
```diff
@@ -123,21 +124,20 @@ private void installAuthorizableConfigurationBean(final Session session,
...
 
-                handleIntermediatePath(session, authorizableConfigBean, history,
-                        authorizableInstallationHistory, userManager);
-                mergeGroup(history, authorizableInstallationHistory,
-                        authorizableConfigBean, userManager);
+            mergeGroup(history, authorizableInstallationHistory,
+                    authorizableConfigBean, userManager);
 
-            } else {
-                String msg = "- authorizable " + principalId
```
ghenzler: This message is removed now - users can be moved now (and references to it from other groups are maintained).

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/validators/impl/AuthorizableValidatorImpl.java
```diff
@@ -36,16 +37,58 @@ public void validate(final AuthorizableConfigBean authorizableConfigBean)
...
-            return validateMemberOf(this.authorizableConfigBean)
+            return validateAuthorizableProperties(this.authorizableConfigBean)
+                    && validateMemberOf(this.authorizableConfigBean)
                     && validateMembers(this.authorizableConfigBean)
                     && validateAuthorizableId(this.authorizableConfigBean);
         }
         return true;
     }
 
+    public boolean validateAuthorizableProperties(
```
ghenzler: Introduced validation for some requirements to fail early

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableutils/impl/AuthorizableCreatorServiceImpl.java
```diff
@@ -662,10 +662,7 @@ private Authorizable createNewUser(
         } else {
             newUser = userManager.createUser(principalId, password, new PrincipalImpl(principalId), intermediatePath);
         }
-
-        if (!isSystemUser) {
```
ghenzler: There is no reason to not run setAuthorizableName() for system users - the current code did run this method on second run anyway (second run=when the system user already exists)

PR #79, opened 2016-05-09
expand import of packages com.day.cq.commons and com.day.cq.commons.j…

santiagozky:
expand import of packages com.day.cq.commons and com.day.cq.commons.jcr for aem 6.2

### DIFF
--- a/accesscontroltool-bundle/pom.xml
+++ b/accesscontroltool-bundle/pom.xml
@@ -168,7 +168,7 @@
               <artifactId>maven-bundle-plugin</artifactId>
               <configuration>
                 <instructions>
-                    <Import-Package>com.day.cq.commons;version="[5.6,6.0]",com.day.cq.commons.jcr;version="[5.6,6.0]",*</Import-Package>
+                    <Import-Package>com.day.cq.commons;version="[5.6,7)",com.day.cq.commons.jcr;version="[5.6,7)",*</Import-Package>
                     <Embed-Dependency>el-api,jasper-el</Embed-Dependency>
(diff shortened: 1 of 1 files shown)


kwin:
Automatically expanding the import range is very dangerous, as Adobe is following semantic versioning, therefore there must be a reason why the major version on those packages has been increased. At least some classes in those packages must have been modified in a backward-incompatible way. I would rather recommend to get rid of those dependencies completely.

ghenzler:
@kwin This is basically about the methods Externalizer.authorLink() and JcrUtil.copy(). Externalizer.authorLink() could be removed fairly easily, but for JcrUtil.copy() it probably would require to copy the code over.

kwin:
Can't we use https://sling.apache.org/documentation/the-sling-engine/sling-api-crud-support.html#copy-a-resource for the latter?

ghenzler:
JcrUtil.copy() copies a deep tree, not only a resource - we could copy the relevant parts of JcrUtil into our namespace (e.g. helper package), but I would prefer to use the original for now. 

Reflecting a bit about semantic versioning and what we are doing here: It is not really only about the AEM dependency here -  let's say jackrabbit-api is bumped to 3.0 some time in the future, then we would either have to maintain two versions of the AC Tool in parallel or allow a "major version spanning" package import and code, that deals with API differences via reflection if needed (already done like that for system user creation). Boils down to a) it's good to minimize dependencies as much as possible, but b) if an expanded version range is required, it's still better than maintaining two versions of the AC Tool.

Issue #99, opened 2016-06-20
Make deletion of memberships to not configured groups configurable


nc-andreashaller:
In 
`biz.netcentric.cq.tools.actool.authorizableutils.impl.AuthorizableCreatorServiceImpl.mergeMemberOfGroupsFromRepo(String, UserManager, Set<String>)`
and
`biz.netcentric.cq.tools.actool.authorizableutils.impl.AuthorizableCreatorServiceImpl.mergeMultipleMembersOfBothGroups(String, AcInstallationHistoryPojo, UserManager, Set<String>, Set<String>)`
all memberships of a configured group to a not configured group (`config-group.isMember(nonconfig-group)`) get deleted.

Generally this makes sense as with such memberships permissions could be 'injected' into the configuration which are not expected and controllable. On the other hand in some cases it is required to not delete those memberships. For example groups which where created by code and or user action:

*UseCase:*
- a group "MyProject Admins" is automatically created by the user action "create project" 
- also automatically the group "All Project Admins" gets the new group "MyProject Admins" as a member

If the group "All Project Admins" gets installed via the ACTool this membership is automatically removed. Also the group "MyProject Admins" cannot be added to the ACTool configuration because it can be dynamically created and deleted by users.

*Solution:*
The ACTool can be configured to check a regular expression before deleting the memberships. If the regular expression matches the membership it will not be deleted.

Issue #40, opened 2015-09-02
Allow YAML files to require a minimum version of the ACTool


kwin:
This is necessary to prevent the YAML files to be installed with older versions having known bugs or with older versions not supporting that YAML format (i.e. because a feature has been added in a certain version).

PR #102, opened 2016-06-23
#99: added OSGi configuration property to filter deletion of membersh…

nc-andreashaller:
#99 added OSGi configuration property to filter deletion of memberships to "external" groups.


ghenzler:
Let's use https://github.com/Netcentric/accesscontroltool/pull/103/commits to fix both #99 and #40 instead.

Issue #112, opened 2016-10-10
Improve content importer 


ghenzler:
Issues #46 and #71 introduced a way to specify content with doc view xml that is created during ACE installation. The class biz.netcentric.cq.tools.actool.helper.ContentHelper uses session.importXML() to achieve this. Unfortunately that importer only supports standard JCR doc view compared to the "enhanced docview serialization" that FileVault supports. Most notably, the import of multi-valued properties with the `[val1,val2,val3]` syntax does not work.

It would be good to use the FileVault API [1] to import content with the exact same format as regular content packages use in their xml files.

[1] https://jackrabbit.apache.org/filevault/apidocs/index.html

ghenzler:
Created https://github.com/Netcentric/accesscontroltool/pull/113 - the FileVault API does not seem to have an easy way to import one `.content.xml`. The best way I found is to use a special "one xml file archive" together with the importer.

ghenzler:
Merged https://github.com/Netcentric/accesscontroltool/pull/113

PR #113, opened 2016-10-10
#112 Support "enhanced doc view" format as used by FileVault

ghenzler:
(using FileVault Archive importer instead of session.importXML)


docs/Configuration.md
```diff
@@ -75,17 +71,13 @@ In general it is best practice to not generate regular users by the AC Tool but
...
-     - preferencesContent: allows to provide docview xml that will reset the preferences node to the given structure after each run (optional, since v1.8.2)
-```
+property | comment | required
+--- | --- | ---
+name, description, path, isMemberOf | Work exactly as for groups | optional
+password | The PW for the user. Obviously this is stored in plain text and should only be used for test users | Required for non-system users, otherwise must not be set
+isSystemUser | Create users as system user (AEM 6.1 and later) | optional
+profileContent | Allows to provide [enhanced docview xml](https://jackrabbit.apache.org/filevault/apidocs/org/apache/jackrabbit/vault/util/DocViewProperty.html) that will reset the profile to the given structure after each run (optional, since v1.8.2) | optional
+preferencesContent | Allows to provide [enhanced docview xml](https://jackrabbit.apache.org/filevault/apidocs/org/apache/jackrabbit/vault/util/DocViewProperty.html) that will reset the preferences node to the given structure after each run (optional, since v1.8.2) | optional
```
kwin: I would rather say since 1.9.0, because the enhanced docview is only supported since the upcoming version.

Issue #105, opened 2016-07-26
Provide a Sling health check that ensures that the last run was successful


ghenzler:
See https://sling.apache.org/documentation/bundles/sling-health-check-tool.html, the history in /var/statistics/achistory can be checked.

ghenzler:
Merged https://github.com/Netcentric/accesscontroltool/pull/119

PR #119, opened 2016-10-18
#105 Adding Sling HC to warn if the last installation was not

ghenzler:
successful.


ghenzler:
@kwin b4beaae makes HC dependency optional (this might only be relevant for integration tests, in AEM many foundation bundles don't start without hc.core)

PR #122, opened 2016-10-19
Feature/#110 delete obsolete groups

ghenzler:


ghenzler:
@kwin Had committed f330359bca84a1436226b76c8d5d938f3f6e5d29 with documentation already, but the connection was gone in the train...

docs/AdvancedFeatures.md
```diff
@@ -150,4 +150,15 @@ An exception to this might be dynamic groups created and maintained by authors o
       allowExternalGroupNamesRegEx: external.* # the AC Tool groups can inherit from other external.* groups
 ```
 
+## Automatically purge obsolete groups
+The root element `obsolete_authorizables` can be used to automatically purge authorizables, that are not in use anymore:
```
gruberroland-netcentric: The doc is talking about authorizables and groups. So is it also possible to delete users? If yes then this should be more clear.

Additionally, an example config should be added to the example content package.

ghenzler: Yes it can also delete users as it is using purgeAuthorizables() under the hood...

Issue #134, opened 2016-11-11
StringIndexOutOfBoundsException


chetanyajain:
Getting the exception when I took the dump of all the permissions and re-executed with 1.9.1 version:

```
11.11.2016 20:45:19.577 *ERROR* [qtp601409526-71271] biz.netcentric.cq.tools.actool.aceservice.impl.AceServiceImpl Exception in AceServiceImpl: {}
java.lang.StringIndexOutOfBoundsException: String index out of range: -1
	at java.lang.String.substring(String.java:1931)
	at org.apache.jackrabbit.oak.security.user.UserProvider.getFolderPath(UserProvider.java:333)
	at org.apache.jackrabbit.oak.security.user.UserProvider.createFolderNodes(UserProvider.java:294)
	at org.apache.jackrabbit.oak.security.user.UserProvider.createAuthorizableNode(UserProvider.java:257)
	at org.apache.jackrabbit.oak.security.user.UserProvider.createGroup(UserProvider.java:190)
	at org.apache.jackrabbit.oak.security.user.UserManagerImpl.createGroup(UserManagerImpl.java:215)
	at org.apache.jackrabbit.oak.security.user.UserManagerImpl.createGroup(UserManagerImpl.java:204)
	at org.apache.jackrabbit.oak.jcr.delegate.UserManagerDelegator$13.perform(UserManagerDelegator.java:226)
	at org.apache.jackrabbit.oak.jcr.delegate.UserManagerDelegator$13.perform(UserManagerDelegator.java:222)
	at org.apache.jackrabbit.oak.jcr.delegate.SessionDelegate.perform(SessionDelegate.java:200)
	at org.apache.jackrabbit.oak.jcr.delegate.UserManagerDelegator.createGroup(UserManagerDelegator.java:222)
	at biz.netcentric.cq.tools.actool.authorizableutils.impl.AuthorizableCreatorServiceImpl.createNewGroup(AuthorizableCreatorServiceImpl.java:495)
	at biz.netcentric.cq.tools.actool.authorizableutils.impl.AuthorizableCreatorServiceImpl.createNewAuthorizable(AuthorizableCreatorServiceImpl.java:362)
	at biz.netcentric.cq.tools.actool.authorizableutils.impl.AuthorizableCreatorServiceImpl.handleIntermediatePath(AuthorizableCreatorServiceImpl.java:256)
[...]

PR #135, opened 2016-11-18
#134 fix index out of bound

jochenkoschorke:


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableutils/AuthorizableCreatorException.java
```diff
@@ -9,7 +9,11 @@
 package biz.netcentric.cq.tools.actool.authorizableutils;
 
 public class AuthorizableCreatorException extends Exception {
-    public AuthorizableCreatorException(String message) {
+    public AuthorizableCreatorException(final String message) {
```
ghenzler: @jochenkoschorke Let's not add `final`s if not explicitly needed

Issue #127, opened 2016-10-21
Support partial update of ACEs (path-based)


ghenzler:
Today on we have an API (and various other means like JMX or install hook) to install a complete setup for a platform instance. Sometimes it can be useful to reapply the ACEs for certain content regions only (e.g. /content/myproj) instead of the complete setup (that would create users, reapply ACEs that rarely change for functionality in /libs, etc.). Applying the whole setup is usually not a problem (in some projects we do it nightly and not only after/during deployment) but performance-wise it can become a problem, if ACEs from the config should be applied straight after an author has created content (usually this is not needed and there are other ways around it, but for some cases it could be useful).

The following API method can be added to `AceService`:

```
void execute(String[] basePaths); // see [1]
void execute(); // <-- exists already
```

Doing an incremental update should 
- fail early if groups are missing (should not happen if a full update was performed after deployment)
- will only delete obsolete ACE entries within the base path regions

[1]
not using `String...` because this could be limiting in the future, e.g. the signature `void execute(String configPath)` would be good to be usable for the future

Issue #146, opened 2017-01-30
Avoid concurrently open session during AC Installation


ghenzler:
Two sessions are created by the tool, one for creating/updating users ("session authorizables"), one for creating/updating ACEs ("session ACEs").  There are some sessions for e.g. saving the history in the end, but they are short-living and not significant in the big picture. 

Currently the "session ACEs" is already created at the very beginning and idles while "session authorizables" is used to create the authorizables. This should be improved to be sequential, meaning "session ACEs"  should only be created after "session authorizables" is saved and closed.

Also see https://github.com/Netcentric/accesscontroltool/blob/1.9.1/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/aceservice/impl/AceServiceImpl.java#L117

Issue #145, opened 2017-01-30
OAK MERGE ERROR when saving the history to "/var/statistics/achistory" for really large AC setups


ghenzler:
It turns out that for large installations the following error can occur:

```
26.01.2017 10:48:53.349 *ERROR* [qtp1594121387-23585-ACTool-Config-Worker] biz.netcentric.cq.tools.actool.installationhistory.impl.AcHistoryServiceImpl RepositoryException: javax.jcr.InvalidItemStateException: OakMerge0001: OakMerge0001: Failed to merge changes to the underlying store (retries 5, 107976 ms)
...
Caused by: org.apache.jackrabbit.oak.api.CommitFailedException: OakMerge0001: OakMerge0001: Failed to merge changes to the underlying store (retries 5, 107976 ms)
...
Caused by: org.apache.jackrabbit.oak.plugins.document.DocumentStoreException: DBObject of size 16962907 is over Max BSON size 16777216
...
Caused by: com.mongodb.MongoInternalException: DBObject of size 16962907 is over Max BSON size 16777216
```

Either we should save as binary or choose to split across multiple properties (the first option probably being the better)

@jochenkoschorke FYI

ghenzler:
Fixed by using the non-verbose log to write the property if the text becomes too big.

PR #148, opened 2017-02-06
Feature/127 support partial update restricted to given paths

ghenzler:


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/aceservice/AceService.java
```diff
@@ -62,13 +70,13 @@
 
     /** Common entry point for JMX and install hook.
      * 
-     * @param session
      * @param history
      * @param configurationFileContentsByFilename
      * @param authorizableInstallationHistorySet
+     * @param restrictedToPaths only apply ACLs to root paths as given
      * @throws Exception */
-    public void installConfigurationFiles(Session session, AcInstallationHistoryPojo history,
```
kwin: I still want to be able to give a dedicated session to the AcService, why has that been removed? Even with intermediate saves there should only be one session used IMHO (and that one should be given as parameter).

ghenzler: Installation can take long - passing this session can mean it has to be open for a really long time (and then you have the concurrent session for hte authorizables). See https://github.com/Netcentric/accesscontroltool/issues/146 for details

ghenzler:
This PR fixes
#127
#145
#146
#149

Issue #153, opened 2017-02-27
Allow to set variables


ghenzler:
In the context of #140 solution option 3 and to avoid duplication, it would be really good to be able to set global variables/constants as follows:

```
- group_config:

    - DEF ldapGroupTemplate="cn=PLACEHOLDER,ou=mydepart,ou=Groups,dc=comp,dc=com;IDPNAME"

    - aem-only-editor-group:
        - name: "AEM Role Editor"
          externalId: ${replace(ldapGroupTemplate, "PLACEHOLDER", "group-name")}

```

Issue #140, opened 2017-01-23
LDAP Group sync issue with AEM 6.2


waringme:
1. clean instance on aem 6.2 SP 1 
2. created groups with netcentric tool and deployed. for example there is a group -AEM_SYSYTEM_ADMIN
3. same groups setup in ldap AEM_SYSYTEM_ADMIN.
4. user1 is in ldap and is a member of AEM_SYSYTEM_ADMIN.
5. user1 get authenticated but fails to get authorisation.  AEM sees the AEM_SYSYTEM_ADMIN as a local group.
6. Deleted the local AEM_SYSYTEM_ADMIN.
7. user1 logs in again and the group from ldap is synced acrossed. because the groups has already had already been run in step 2. false positive.
8. delete the synced groups from crx.
9. installed the netcentric group package again so the AEM_SYSYTEM_ADMIN is created.
10. logged in again with user1.
11. netcentric group AEM_SYSYTEM_ADMIN is now showing as synced with ldap and displaying the members (user 1 etc)

Is there a way to controlling permissions for LDAP groups? 

Thanks

Tim

ghenzler:
@waringme Can you check in more detail what the group AEM_SYSTEM_ADMIN looks like in JCR when created by AC Tool vs. created by LDAP sync? I could imagine that a flag was introduced with SP1 (SAML has such a flag for external groups). Also, can you please provide your oak version? (current oak hotfixes have changed LDAP behaviour, SP1 has oak 1.4.6 [2])

[1] https://helpx.adobe.com/experience-manager/kb/aem62-available-hotfixes/OakCumulativeFixPack.html#Oak147
[2] https://docs.adobe.com/content/docs/en/aem/6-2/release-notes/sp1.html#contentbody_text_2013062266

waringme:
I've not got access to the system at the moment.  from memory the external groups nodes looked very similar to 6.1. I can check next Monday. 
Version:  oak 1.4.9.
Server setup was done by
1. AEM 6.2 GA
2. Hotfix 12785
3. SP 1
4 .Hot fixes 12444, 13062,13547

ghenzler:
The root cause for it is really the change in OAK-4397 (change active since oak 1.4.7), possible workarounds are described in [1]. 

For AC Tool users this means the easiest workaround is to always run ldap sync via JMX *before* running the AC Tool (hence new groups would always be created by the LDAP module and therefore the problem should not show up). 

[...]

PR #154, opened 2017-02-28
#140 Adding config attribute 'externalId'

ghenzler:
in order to allow to precreate LDAP groups. Also adding ability to use 
variables (#153)

### DIFF
--- a/accesscontroltool-bundle/pom.xml
+++ b/accesscontroltool-bundle/pom.xml
@@ -149,6 +149,13 @@
         </dependency> 
         
         <dependency>
+          <groupId>org.apache.jackrabbit</groupId>
+          <artifactId>oak-auth-external</artifactId>
+          <version>1.2.7</version>
+          <optional>true</optional>
+        </dependency> 
+            
+        <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-log4j12</artifactId>
             <version>1.5.10</version>
@@ -190,6 +197,12 @@
               <configuration>
                 <instructions>
                     <Embed-Dependency>el-api,jasper-el</Embed-Dependency>
+                    <!-- Required to support 6.1 and 6.2 -->
+                    <Import-Package>
+                        org.apache.jackrabbit.oak.spi.security.authentication.external;version="[1.0,3)",
+                        *
+                    </Import-Package>              
+                    
                 </instructions>
                 
               </configuration>
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableutils/impl/ExternalGroupCreatorServiceImpl.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableutils/impl/ExternalGroupCreatorServiceImpl.java
@@ -1,0 +1,128 @@
+/*
+ * (C) Copyright 2017 Netcentric AG.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package biz.netcentric.cq.tools.actool.authorizableutils.impl;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.Set;
+
+import javax.jcr.RepositoryException;
+import javax.jcr.Session;
+import javax.jcr.ValueFactory;
+
+import org.apache.commons.lang.StringUtils;
+import org.apache.felix.scr.annotations.Component;
+import org.apache.felix.scr.annotations.Service;
+import org.apache.jackrabbit.api.security.user.Authorizable;
+import org.apache.jackrabbit.api.security.user.AuthorizableExistsException;
+import org.apache.jackrabbit.api.security.user.Group;
+import org.apache.jackrabbit.api.security.user.UserManager;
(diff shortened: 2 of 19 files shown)


accesscontroltool-bundle/pom.xml
```diff
@@ -190,6 +197,12 @@
               <configuration>
                 <instructions>
                     <Embed-Dependency>el-api,jasper-el</Embed-Dependency>
+                    <!-- Required to support 6.1 and 6.2 -->
+                    <Import-Package>
+                        org.apache.jackrabbit.oak.spi.security.authentication.external;version="[1.0,3)",
```
kwin: really dangerous.

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableutils/impl/ExternalGroupCreatorServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+
+import biz.netcentric.cq.tools.actool.authorizableutils.AuthorizableCreatorException;
+import biz.netcentric.cq.tools.actool.authorizableutils.AuthorizableInstallationHistory;
+import biz.netcentric.cq.tools.actool.configmodel.AuthorizableConfigBean;
+import biz.netcentric.cq.tools.actool.helper.Constants;
+import biz.netcentric.cq.tools.actool.installationhistory.AcInstallationHistoryPojo;
+
+@Service(ExternalGroupCreatorServiceImpl.class)
+@Component(metatype = false)
+public class ExternalGroupCreatorServiceImpl {
```
kwin: Can you add some javadoc that this service might not be started as well as a rough description

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableutils/impl/ExternalGroupCreatorServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(ExternalGroupCreatorServiceImpl.class);
+
+    public Authorizable createGroupWithExternalId(
+            final UserManager userManager,
+            final AuthorizableConfigBean authorizableConfigBean,
+            AcInstallationHistoryPojo status,
+            AuthorizableInstallationHistory authorizableInstallationHistory,
+            ValueFactory vf,
+            Map<String, Set<AuthorizableConfigBean>> principalMapFromConfig, Session session)
```
kwin: why session, status and authorizableInstallationHistory as parameter?

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableutils/impl/ExternalGroupCreatorServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        @Override
+        public ExternalIdentityRef getExternalId() {
+            return ExternalIdentityRef.fromString(authorizableConfigBean.getExternalId());
+        }
+
+        @Override
+        public Map<String, ?> getProperties() {
+            return Collections.<String, Object> emptyMap();
```
kwin: isn't it safer to just throw http://docs.oracle.com/javase/8/docs/api/?java/lang/UnsupportedOperationException.html for all methods which are not supposed to be used?

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableutils/impl/ExternalGroupCreatorServiceImpl.java
kwin: Also wouldn't it be easier to just rely on `ExternalIdentityRef.fromString` to create the group instead of implementing DefaultSyncContext and ExternalGroup?

accesscontroltool-bundle/pom.xml
ghenzler: The only way to get it to work for AEM 6.1 and 6.2. The minimal parts of the package it is using should really be fairly stable (in fact it was really annoying to find out that the only reason it did not work was the OSGi version!)

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableutils/impl/ExternalGroupCreatorServiceImpl.java
ghenzler: I thought about this as well, but (hopefully) it is better/safer to rely on this extension point that is explicitly given for external identity providers (like e.g. oak-auth-ldap), especially if the way groups are created changes slightly in the future.

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableutils/impl/ExternalGroupCreatorServiceImpl.java
ghenzler: 50/50 chance of what will be better in the future - throwing the UnsupportedOperationException also could cause a problem where actually no problem is in some future versions. I think it's better to provide a "quiet NOOP implementation" for this case.

Issue #161, opened 2017-03-10
Performance problem in AcHelper.getPrincipalForName 


ghenzler:
The use of userManager.getAuthorizable(String principalId) in AcHelper.getPrincipalForName() as huge performance impact - using userManager.getAuthorizable(Principal principalId) instead brings down execution time for an example with 16816 ACLs from 5 hours to 3min (100x faster).

PR #162, opened 2017-03-10
#161 Using userManager.getAuthorizable(Principal principalId) instead of userManager.getAuthorizable(String principalId)

ghenzler:


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/AcHelper.java
```diff
@@ -135,7 +136,7 @@ private static Principal getPrincipalForName(final Session session, String princ
         // Also see https://issues.apache.org/jira/browse/OAK-3228
         final JackrabbitSession js = (JackrabbitSession) session;
         final UserManager userManager = js.getUserManager();
-        final Authorizable authorizable = userManager.getAuthorizable(principalName);
+        final Authorizable authorizable = userManager.getAuthorizable(new PrincipalImpl(principalName));
```
kwin: Is this better, because the name is actually the id, so the lookup is much faster?

Issue #167, opened 2017-03-23
Fix inconsistencies in purge authorizables logic when using external ids


ghenzler:
the JMX function purgeAuthorizables and the `obsolete_authorizables` configuration does currently not work with groups that have an externalId set.

ghenzler:
Fixed on both `develop_19x` and `develop`

PR #169, opened 2017-03-23
#167 Fixed purging of authorizables

ghenzler:


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/QueryHelper.java
```diff
@@ -30,202 +32,213 @@
...
+        Set<Node> nodeSet = new LinkedHashSet<Node>();
+
+        Iterator<String> authorizablesIdIterator = authorizableIds.iterator();
+
+
+        while (authorizablesIdIterator.hasNext()) {
+            StringBuilder queryStringBuilder = new StringBuilder();
+            queryStringBuilder.append(
+                    "SELECT ace.* FROM [rep:ACE] AS ace INNER JOIN [rep:Authorizable] AS authorizable "
+                            + "ON ace.[rep:principalName] = authorizable.[rep:principalName] WHERE ");
```
ghenzler: So here we have to do a join. Performance of it is better than I expected (did some test in the groovy console that made the difference almost neglectable).

Issue #163, opened 2017-03-17
Error when executing the hook under AEM 6.3


santiagozky:
When executing the installhook under AEM 6.3 the following error happens

`Caused by: javax.jcr.LoginException: Bundle biz.netcentric.cq.tools.accesscontroltool.bundle is NOT whitelisted
	at org.apache.sling.jcr.base.AbstractSlingRepository2.loginAdministrative(AbstractSlingRepository2.java:378)
	at biz.netcentric.cq.tools.actool.dumpservice.impl.DumpserviceImpl.createAclDumpMap(DumpserviceImpl.java:419)`

This seems to be caused by the usage of repository.loginAdministrative. which now is not allowed unless the bundle calling it is explicitly whitelisted  in org.apache.sling.jcr.base.internal.LoginAdminWhitelist

kwin:
There are three solution approaches to this:
1. Whitelist the usage of of the administrative session (https://issues.apache.org/jira/browse/SLING-5135)
2. Use a dedicated service user for this. Not clear how to initially create it with the appropriate rights or whether an existing service user should be just reused.
3. Use the session of the underlying package installation

3. is IMHO the best option for the Vault Install Hook, but not an option for JMX as there is no session/resourceresolver available there. So for JMX we should instead rely on 2.

santiagozky:
I've created a PR with a proporsal for this issue.
https://github.com/Netcentric/accesscontroltool/pull/177
it provides a service user, then it is easy to use loginService instead of loginAdministrative.

ghenzler:
Fixed by #177, after some tests removed cloning of session in 821c4fa6d0a878f81d173f546d618982d7c88a19 to accommodate @kwin's suggestion.

PR #170, opened 2017-03-25
#163 Fixing Error when executing the hook under AEM 6.3

karolisma:
Added newer sling api version.

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/session/impl/SessionManagerImpl.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/session/impl/SessionManagerImpl.java
@@ -1,0 +1,104 @@
+package biz.netcentric.cq.tools.actool.session.impl;
+
+import biz.netcentric.cq.tools.actool.session.SessionManager;
+import org.apache.felix.scr.annotations.Activate;
+import org.apache.felix.scr.annotations.Component;
+import org.apache.felix.scr.annotations.Reference;
+import org.apache.felix.scr.annotations.Service;
+import org.apache.sling.api.resource.LoginException;
+import org.apache.sling.api.resource.ResourceResolver;
+import org.apache.sling.api.resource.ResourceResolverFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.jcr.RepositoryException;
+import javax.jcr.Session;
+import java.util.HashMap;
+import java.util.Map;
+
+@Component
+@Service
+public class SessionManagerImpl implements SessionManager {
+
+    private static final Logger LOG = LoggerFactory.getLogger(SessionManagerImpl.class);
+
+    ThreadLocal<Session> sessionHolder;
+
+    @Reference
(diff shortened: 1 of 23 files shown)


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/session/impl/SessionManagerImpl.java
```diff
@@ -0,0 +1,104 @@
...
+        if(session != null) {
+            return session;
+        }
+
+        return getServiceUserSession(session);
+    }
+
+    @Override
+    public void close(Session session) {
+        if(session.equals(sessionHolder.get())) {
```
santiagozky: maybe instead we should wrap the session in a custom implementation that prevents closing it? that way a developer cannot accidentally just call session.logout()

kwin:
I would very much be in favor of passing a session to be used as parameter. That way the caller (i.e. JMX or Vault Hook) can explicitly determine which session/service user to use. Hiding that in a kind of magic service (using either service resource resolvers or thread resource resolvers) is too magic for me. Also it is not clear to me, how would close the session/resource resolver then?

We should make sure, that the session is closed as soon as possible (and not being reused among multiple threads)

ghenzler:
I agree with @kwin that we should reuse credentials (but not sessions) as much as possible. This means for me
- For the install hook case, the package manager session should be cloned for all spots where currently the admin session is opened https://java.net/jira/browse/JSR_333-27
- For the plain JMX case (not via /system/console/jmx) there is definitely now session we can rely on. For that we probably just have to create a new session 
- For the System Console JMX case, it is probably wise to use the sling request session for cloning (@kwin WDYT? better than creating a new session with a system user configured IMHO)

Generally I would not like to change anything regarding session lifetimes and session count ATM - I would try to clone sessions from existing sessions if possible (and only create from scratch if really necessary)

kwin:
> For the System Console JMX case, it is probably wise to use the sling request session for cloning (@kwin WDYT? better than creating a new session with a system user configured IMHO)

JMX has never access to a Sling Request. Even if accessed through the Felix Web Console, this request is not handled by Sling, therefore only a HttpServletRequest is available which has no `ResourceResolver` or `Session`.

ghenzler:
@kwin You don't have access to the Sling Request, but that is not needed. I could image that [1] does work (albeit not a very clean solution) - @karolisma have you tested if you actually can get a session like this? (and if yes, does it still work if you use plain curl to trigger it?)

[1]
https://github.com/karolisma/accesscontroltool/blob/9c0f02ee9682a09019e6d253704ae641ebac8176/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/session/impl/SessionManagerImpl.java#L70

kwin:
> I could image that [1] does work (albeit not a very clean solution) 

`resourceResolverFactory.getThreadResourceResolver();` returns null for Non-Sling threads/requests.

karolisma:
@ghenzler I have tested, it works on installer hook and via jmx (/system/console), if user logged in has permissions needed.
I haven't tried it via curl.

Issue #151, opened 2017-02-23
If no ACEs remain for a path, the relating empty rep:policy node should be deleted as well


ghenzler:
For the edge case that an older version of the AC configuration contained a path and a newer version deletes all ACEs, currently the tool leaves an empty "rep:policy" (without children). This is similar to what remains in CRXDE if you remove all ACEs, but you do not delete the ACL for the path (two clicks needed in CRXDE as well). However the ACTool should detect that no ACEs are left for a path and remove the rep:policy node as well (using `AccessControlManager.removePolicy(String, AccessControlPolicy)`)

ghenzler:
Fixed on both `develop_19x` and `develop`

PR #173, opened 2017-03-29
#151 For the case in acl is empty, remove the rep policy node

ghenzler:


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/AccessControlUtils.java
```diff
@@ -153,58 +153,6 @@ public static JackrabbitAccessControlList getAccessControlList(
         return null;
     }
 
-    /** A utility method to add a new access control entry.<br>
```
ghenzler: This was just dead code

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/acls/AceBeanInstallerImpl.java
```diff
@@ -291,7 +291,11 @@ private void install(AceBean aceBean, final Session session, Principal principal
                     + ", principal: " + principal.getName() + ", privileges: "
                     + aceBean.getPrivilegesString() + ", allow: " + aceBean.isAllow());
         }
-        acMgr.setPolicy(aceBean.getJcrPath(), acl);
+        if (!acl.isEmpty()) {
```
ghenzler: only use setPolicy() if not empty, otherwise remove

PR #177, opened 2017-03-30
replace loginAdministrative with loginService with a provided user

santiagozky:


accesscontroltool-package/src/main/META-INF/vault/filter.xml
```diff
@@ -1,4 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <workspaceFilter version="1.0">
-	<filter root="/apps/netcentric/actool"/>       
+	<filter root="/apps/netcentric/actool"/>  
+	<filter root="/home/users/system/netcentric/actool/acl/ynUbs6nnylcpRBs0eXdv"/>
```
ghenzler: I would use the path /home/users/system/actool/actool-service (if you do not use the api to create a user but a package, you can use an unobfuscated path)

accesscontroltool-package/src/main/META-INF/vault/filter.xml
```diff
@@ -1,4 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <workspaceFilter version="1.0">
-	<filter root="/apps/netcentric/actool"/>       
+	<filter root="/apps/netcentric/actool"/>  
+	<filter root="/home/users/system/netcentric/actool/acl/ynUbs6nnylcpRBs0eXdv"/>
+	<filter root="/" mode="merge"/>
```
ghenzler: This potentially can take veeery long to install (depending on how much content the instance has). Do we have a chance to set it to /rep:policy or similar? If not, the ACE will have to be added by a an install hook (custom code) or repo init (or whatever else can do it)

ghenzler:
@kwin For the JMX use case we already need the service user. KISS is to just use it for the install hook case as well. I agree that we should create fewer sessions, but actually it is not as many as it seems (many of the sessions are created for special operations like dump acls that are only available via jmx/api)

kwin:
> KISS is to just use it for the install hook case as well. 

This is a security issue, because then even an unprivileged user (with only read/write access to /etc/packages) can destroy ACEs by just installing a forged package

accesscontroltool-package/src/main/META-INF/vault/filter.xml
santiagozky: thanks. I'll have a look

ghenzler:
> This is a security issue, because then even an unprivileged user can destroy ACEs by just installing a forged package
You are right, I didn't consider this -> we will have to deal with the two situations.

gruberroland-netcentric:
I would not consider write access to /etc/packages access as "unprivileged".

santiagozky:
well. I made some changes.
now session is opened at the beginning of the process (jmx, event, etc) and passed as parameter.
it still saves when it did before, but I close the session at the end.
I don’t think it makes sense to pass the session around in some cases and in others open it at the middle of the process. sessions live a little longer, but code is more uniform.

accesscontroltool-package/src/main/META-INF/vault/filter.xml
ghenzler: @santiagozky I don't think we can leave it like this - this will traverse the whole content tree and hence take way to long on production.

accesscontroltool-package/src/main/META-INF/vault/filter.xml
ghenzler: Also this is still using the cryptic name. It should use the name "actool-service" or similar (in the same way as e.g. "authentication-service")

ghenzler:
@santiagozky I think we should change a lot fewer method signatures. Most places can just use getServiceUser instead of getting the admin user. Only the call stack that passes AceServiceImpl.installAcConfiguration() is involved in actually installing ACEs (and can be triggered by package installation), all other places are only triggered by JMX and hence can use the place of instantiating a session.

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/aceservicejmx/impl/AceServiceMBeanImpl.java
```diff
@@ -53,41 +57,126 @@ public AceServiceMBeanImpl() throws NotCompliantMBeanException {
...
+    SlingRepository repository;
+
     @Override
     public String apply() {
-        return aceService.execute().toString();
+        Session session = null;
+        try {
+            session = repository.loginService(Constants.USER_AC_SERVICE, null);
+            return aceService.execute(session).toString();
+        } catch (RepositoryException e) {
```
kwin: I would rather rely on the exception handling of the code calling the mbean. So just throwing here is fine.

santiagozky:
@ghenzler  a session is also needed to obtain the configuration (since you can evaluate jcr nodes in there). Wouldn't it cause problems to obtain that using an admin session and then trying to write it using a session with possibly different set of permissions?

PR #188, opened 2017-05-02
Allow spaces in authorizable ids

YegorKozlov:
Please allow spaces in group ids. The GROUP_ID_PATTERN regex is too strict.

I agree that spaces _should_ not be used, but in our setup group ids come from an external system and spaces are common. Also, OAK's UserManager#createGroup(String groupId) can create group ids with spaces just fine.


ghenzler:
We fully agree that it is not nice :)  - but you have a point, if it works and is valid we have to support it.

Issue #189, opened 2017-05-04
Allow to set distinct principal names for users/groups being created


kwin:
Currently AC Tools always assumes that the authorizable id is equal to the principal name. Jackrabbit does not enforce this equality. In fact Oak uses different authorizable ids and principal names for groups created from the LDAP. Therefore I would propose the following changes:

1. Clarify in the documentation that the entry value for the group configuration (https://github.com/Netcentric/accesscontroltool/blob/develop/docs/Configuration.md#configuration-of-groups) and user configuration (https://github.com/Netcentric/accesscontroltool/blob/develop/docs/Configuration.md#configuration-of-users) refers to authorizable ids. 
2. Clarify that the ACE section (https://github.com/Netcentric/accesscontroltool/blob/develop/docs/Configuration.md#configuration-of-aces) refers to principal names (which are usually equal to the authorizable ids)
2. Optionally allow to specify a dedicated `rep:principalName` property for both groups and users and to set that accordingly when the YAML is processed.

ghenzler:
Points 1. and 2. are done (documentation). 

@kwin Can you think of a real-world use case for 3.?  (LDAP and other externalIds are covered already by `externalId` setting....)

PR #190, opened 2017-05-04
clarify user/group id and principal name

kwin:
As outlined in #189 we should first clarify in the documentation what is being used as identifier for ACEs and users/groups.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -30,11 +30,11 @@ Examples:
 
 <img src="images/configuration-file-structure.png">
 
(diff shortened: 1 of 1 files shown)


docs/Configuration.md
```diff
@@ -108,7 +108,9 @@ Group memberships can be set on user entry or group entry or both.
 
 ## Configuration of ACEs
 
-The configurations are done per principal followed by indented settings for each ACE. This data includes
+The configurations are done per principal (given through its name) followed by indented settings for each ACE. Currently only principal names are supported which are equal to one of the created group ids.
```
ghenzler: @kwin this is not entirely true - while usually principalName=authorizableId when using the AC Tool, for the case the externalId (think LDAP) is used, the section `ace_config` has to list the authorizableId (while the tool sets automatically and correctly the corresponding principalName to the content paths)

ghenzler: Improved doc in 0209e7cd44d52d087ea0ea351e342b511b6c9038

Issue #212, opened 2017-07-20
Add 'unmanagedAcePathsRegex' for authorizables


jon-entwistle:
ACEs added for groups via the web console are overwritten when accesstool is applied, requiring manual intervention to add these permissions again via the web console. 

The proposed feature allows paths to be configured at group level where existing privileges are respected when accesstool is run. These paths need to be recursive, so any ACEs for the group on subpaths are also respected.

An example group configuration would be:

    - A_GRP:
       - name: aName
         memberOf: anOther
         path: u
         honorPrivilege:
           - /respected/path/a
           - /respected/path/b

ghenzler:
@jon-entwistle what about calling the property `restrictAcesToPaths`? That way the following could work (trying to give a more relevant example where it actually makes sense, most of the time we would exactly avoid such a path restriction as this can lead to strange effects):

```
- everyone:
   - name: Everyone
     restrictAcesToPaths:
       - /content/myproj
       - /content/dam/myproj
```
The above example would not touch the other out of the box ACEs for everyone, it would only add (or remove obsolete ones) the ones under the project root paths.

@jon-entwistle do you have maybe even a better example?

jon-entwistle:
@ghenzler the name restrictAcesToPaths would imply that only ACEs which are on the given paths would be applied. This is not the case.

What honorPrivilege does is to find any ACEs which have been defined for the group prior to running the tool, and ensures that these are applied after the tool is run. There should be no side effects. It has exactly the same effect as running the tool (this resetting the permissions) then applying the ACEs which are not encapsulated by the .yaml by hand via the console. 

Hence the name, as any ACEs defined for the given group for the given paths will be honoured.

ghenzler:
@jon-entwistle as discussed offline the "honor privileges" use case can be achieved differently (using https://github.com/Netcentric/accesscontroltool/blob/develop/docs/AdvancedFeatures.md#configure-memberships-to-dynamic-groups). But I think it would be reasonable to have a `restrictAcesToPaths` feature to allow ootb groups to be redefined without having to repeat all ACEs that come ootb - that way it would be easy to just add a few ACEs for e.g. `everyone` to certain repo locations without touching the ootb ACEs (see example in https://github.com/Netcentric/accesscontroltool/issues/212#issuecomment-316667053)

inedoviesov1:
One more use case related to requested feature.

[...]

PR #213, opened 2017-07-20
Feature/212 honor privilege

jon-entwistle:
Adds the capability to define honour privileges at group level: 

Recursive ACEs for the group(s) at the given honour path(s) are stored before accesstool resets privileges to the .yaml definitions. The stored ACEs are then re-applied for the given group(s)/path(s) after the reset is complete.


ghenzler:
@jon-entwistle Doing it like this (as wrapper around the whole AC Tool execution) is a very unclean approach IMHO. 

Two points:
* First we need to find a valid use case where this makes sense
* Then we need to find an implementation that does not act as a wrapper but is respected when each individual path is processed.

ghenzler:
Let's discuss more in #212 on how to do this in reasonable way.

PR #235, opened 2017-11-20
Added colon to regex

Pekket:
I think a fix for

> validation error while reading group data: invalid group name: TEST-AUTHOR:123455


kwin:
this is obsoleted by PR #238.

Issue #251, opened 2017-12-13
Allow to specify encrypted passwords 


ghenzler:
Usually we use plain text passwords for users - this is not really a problem for test users. For some rare cases it is necessary to specify a password for a non-system, production user (most notably a "replication receiver user" for publish). This should be possible using the same semantics as we have them in JCR: If a password value is enclosed in brackets `{encryptedPWStr}`, then it should be automatically be decrypted using com.adobe.granite.crypto.CryptoSupport.

PR #252, opened 2017-12-16
#251: use CryptoSupport for passwords enclosed in brackets.

otarsko:
Tested on AEM 6.1 SP1 and AEM 6.3 via JMX apply method.

### DIFF
--- a/accesscontroltool-bundle/src/test/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthorizableInstallerServiceImplTest.java
+++ b/accesscontroltool-bundle/src/test/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthorizableInstallerServiceImplTest.java
@@ -22,6 +22,7 @@ import static org.mockito.Mockito.verifyNoMoreInteractions;
 import static org.mockito.Mockito.verifyZeroInteractions;
 
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.HashSet;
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -107,6 +107,12 @@
                 <version>5.6.4</version>
                 <scope>provided</scope>
             </dependency>
+            <dependency>
```
otarsko: Note, that it's a version from AEM 6.1 and does not seems to be publicly available.

accesscontroltool-bundle/src/test/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthorizableInstallerServiceImplTest.java
```diff
@@ -40,248 +41,310 @@
...
 
+import com.adobe.granite.crypto.CryptoException;
+import com.adobe.granite.crypto.CryptoSupport;
+
 import biz.netcentric.cq.tools.actool.configmodel.AcConfiguration;
 import biz.netcentric.cq.tools.actool.configmodel.AuthorizableConfigBean;
 import biz.netcentric.cq.tools.actool.configmodel.GlobalConfiguration;
 import biz.netcentric.cq.tools.actool.history.AcInstallationLog;
 
-@RunWith(MockitoJUnitRunner.class)
```
otarsko: Moved all existing tests into nested class to make management of a high number of test easier.

PR #265, opened 2018-02-23
Update dependencies for AEM 6.1

### LINKED ISSUE #240: Upgrade dependencies to AEM 6.1
Although with the ACTool 2.0.0 the minimum requirement was adjusted to AEM 6.1 most of the managed dependencies are a much older version. The depMgmt in the reactor pom should be upgraded to the versions shipped with AEM 6.1 (e.g. OSGi R5, which allows to use generics). Unfortunately we cannot yet rely on the UberJar as only since AEM 6.2 the unobfuscated UberJar is publically available from Adobe's Maven repo.

kwin:
This fixes #240.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -86,37 +86,39 @@
             <dependency>
                 <groupId>org.apache.jackrabbit</groupId>
                 <artifactId>jackrabbit-jcr-commons</artifactId>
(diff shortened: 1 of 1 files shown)


kwin:
This was reverted with e8ce2f9.

PR #273, opened 2018-03-09
Bugfix/240 fix depmgmt for aem61

kwin:
2nd try

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,1 @@
+language: java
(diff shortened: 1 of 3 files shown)


kwin:
@ghenzler This is almost working, but unfortunately the dependency `com.adobe.granite:com.adobe.granite.crypto:3.0.0` (although available in the Adobe Maven Repo, references an unknown parent pom.xml). There is no other suitable version though.... Not sure, why this is not a problem locally.

The other things I am not sure about: {{oak-core}} and {{oak-auth-external}} are both referenced in version 1.2.7 but AEM 6.1 SP1 ships with 1.2.2. Can you explain those?

PR #276, opened 2018-04-11
Evaluate return values of Group.addMember and Group.removeMember

### LINKED ISSUE #275: Check return value of Group.addMember and Group.removeMember and log error accordingly
Currently a failure to add an authorizable as member to an existing group is not properly checked in code. The return value of https://github.com/Netcentric/accesscontroltool/blob/5dd1e91f668457d111f87780362ce49d8b36a3bc/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthorizableInstallerServiceImpl.java#L184 and https://github.com/Netcentric/accesscontroltool/blob/5dd1e91f668457d111f87780362ce49d8b36a3bc/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthorizableInstallerServiceImpl.java#L196 must be properly checked for the return value `true`. Otherwise at least a warning should be issued. 

This is especially useful if someone is trying to add members to group "everyone" which is not supported (compare with https://issues.apache.org/jira/browse/OAK-7323).

kwin:
This closes #275


kwin:
Currently the tests are still failing, unsure why. @ghenzler Maybe you can take a look.

kwin:
@ghenzler I fixed now the tests.

Issue #290, opened 2018-05-16
Regression when using migrateFrom: Fix NPE in removeUmmanagedPrincipals


Amoratinos:
AuthorizablesConfig holds a set of AuthorizableConfigBean and exposes a method to get which principals could be removed from a given path.

When a non-existent name is included in that principal's list the method is throwing a NullPointerException:
```
java.lang.NullPointerException: null
	at biz.netcentric.cq.tools.actool.configmodel.AuthorizablesConfig.removeUnmanagedPrincipalNamesAtPath(AuthorizablesConfig.java:71)
	at biz.netcentric.cq.tools.actool.aceinstaller.BaseAceBeanInstaller.installPathBasedACEs(BaseAceBeanInstaller.java:92)
	at biz.netcentric.cq.tools.actool.impl.AcInstallationServiceImpl.installAces(AcInstallationServiceImpl.java:390)
	at biz.netcentric.cq.tools.actool.impl.AcInstallationServiceImpl.installAcConfiguration(AcInstallationServiceImpl.java:247)
	at biz.netcentric.cq.tools.actool.impl.AcInstallationServiceImpl.installMergedConfigurations(AcInstallationServiceImpl.java:528)
	at biz.netcentric.cq.tools.actool.impl.AcInstallationServiceImpl.installConfigurationFiles(AcInstallationServiceImpl.java:209)
	at biz.netcentric.cq.tools.actool.impl.AcInstallationServiceImpl.apply(AcInstallationServiceImpl.java:165)
	at biz.netcentric.cq.tools.actool.impl.AcInstallationServiceImpl.apply(AcInstallationServiceImpl.java:148)
...
```

PR #292, opened 2018-05-17
Skip getting/managing members of the built-in "everyone" group

stefanseifert:
If one needs to set some ACLs for the everyone group it has to be redefined - example:
```
  - everyone:
    - path: /home/groups/e
      # everything outside the given paths should not be managed by the ac tool
      unmanagedAcePathsRegex: ^(?!(/libs/social/console/content/sites|/libs/cq/core/content/nav/communities/sites)).*$
```
but the everyone group is a special group that has be definition no members (any user is implicitly member of it).

the AC tool handles it as every other group, iterating over all existing members to check if any changes are necessary.

and if a lot of users are installed on the system (e.g. 100,000 users) it takes a considerable amount of time to iterate through all of them, although it makes not sense at all because the members of everyone cannot be changed.

this simple PR avoids this by adding some condition checks. maybe there is more special handling needed for the everyone group, but this change mitigates the described performance problem.

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthorizableInstallerServiceImpl.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthorizableInstallerServiceImpl.java
@@ -99,6 +99,7 @@ public class AuthorizableInstallerServiceImpl implements
             AuthorizableExistsException, AuthorizableCreatorException, CryptoException {
 
         String authorizableId = authorizableConfigBean.getAuthorizableId();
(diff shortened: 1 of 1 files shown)


ghenzler:
Thanks @stefanseifert for finding this issue. As many other actions also do not make sense for principal everyone (moving/migrating), I made it even more restrictive in #293

stefanseifert:
thanks for the quick response & merging!
do you have a plan when the next release can be expected?

ghenzler:
@stefanseifert you were really lucky because I was about to create a release for #290 that we need elsewhere, your fix is included in [2.0.9](https://github.com/Netcentric/accesscontroltool/releases/tag/2.0.9) that I just released.

stefanseifert:
thanks!

PR #296, opened 2018-05-24
Feature/295 update ac tool hook documentation

carlosmorenoin:
Update documentation for projects with the content-package-maven-plugin on versions equals or higher than 1.0.2


carlosmorenoin:
Text updated

PR #309, opened 2018-07-26
#308 also set password for newly created users

### LINKED ISSUE #308: Setting an encrypted password for a new user does not work
Currently, if an encrypted password for a new user is set, it only becomes active during second run of AC Tool because decryption is not done when creating the user.

ghenzler:
Closes #308


ghenzler:
https://github.com/Netcentric/accesscontroltool/pull/310 is better (to not set the pw twice on a new user)

PR #315, opened 2018-08-07
314. Exclude dependencies on osgi-core from bndlib

santiagozky:


santiagozky:
@kwin  right. that seems better. I'll close this PR, then.

Issue #312, opened 2018-08-02
Allow to have conditions for run modes


kwin:
Run modes are often used to distinguish between different aspects of a instance. It is very common to have a run mode per environment (i.e. for DEV/STAGE/PROD). In the case of SSO with the same backend you often want to use different group names for those instances (as otherwise a user having certain rights on a DEV instance automatically has the same rights on PROD). Instead of using runmode specific YAML files it would be handy if you could reference run modes within conditions within the YAML file. That way less code duplication is necessary.

kwin:
This would go nicely along with #311.

ghenzler:
With 022eaeb the syntax would be `containsElement(RUNMODE, 'runmode')` - or should we have a custom syntax for this?

ghenzler:
Changed syntax to `containsItem(RUNMODES, 'runmode')` @kwin's comment https://github.com/Netcentric/accesscontroltool/pull/321#discussion_r224343924

Added unit test to show that `- FOR runmode IN ${RUNMODES}: ` works correctly.

PR #321, opened 2018-10-10
Feature/#311 contains element function to check if string is in array

### LINKED ISSUE #311: Add ArrayUtils#contains to imported functions
It would be great to use https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/ArrayUtils.html#contains-java.lang.Object:A-java.lang.Object- as the rest of imported functions. That would allow for something like
` DEF strings=[hey,ho]`
 `- FOR item IN CHILDREN OF /my/repo:`
 `     - IF ${strings.contains(item.name)}:`

Thanks!

ghenzler:
closes #311 #312

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/configreader/YamlMacroElEvaluator.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/configreader/YamlMacroElEvaluator.java
@@ -12,6 +12,7 @@ import java.beans.FeatureDescriptor;
 import java.lang.reflect.Method;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 
 import javax.el.ArrayELResolver;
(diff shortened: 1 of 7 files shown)


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/configreader/YamlMacroElEvaluator.java
```diff
@@ -99,7 +100,8 @@ public ElFunctionMapper() {
                         StringUtils.class.getMethod("split", new Class<?>[] { String.class, String.class }),
                         StringUtils.class.getMethod("join", new Class<?>[] { Object[].class, String.class }),
                         ArrayUtils.class.getMethod("subarray", new Class<?>[] { Object[].class, int.class, int.class }),
-
+                        YamlMacroElEvaluator.class.getMethod("containsElement", new Class<?>[] { List.class, String.class }),
```
kwin: I would rather go for containsItems(), compare with https://stackoverflow.com/questions/31427117/difference-between-element-item-and-entry-when-it-comes-to-programming

ghenzler: yes, that's better +1

PR #324, opened 2018-10-22
Switch to official OSGi annotations

### LINKED ISSUE #314: The access control tool bundle exposes org.osgi:org.osgi.core:jar:4.3.1 as transitive dependency of bndlib
The access control tool bundle has a dependency on bndlib version 2.3.0 which causes a transitive dependency on org.osgi:org.osgi.core:jar:4.3.1.
This is not a problem for the bundle itself, because it declares dependencies on osgi on version 6.0.0.

The problem is that the scope for bndlib is _compile_ and the osgi-core is _provided_, which means that the bndlib dependencies are propagated to anyone using the bundle as dependency. This causes that if you use the bundle, depending on the order of your dependencies, you will depend on old versions of the annotations.

`[INFO] |  \- biz.netcentric.cq.tools.accesscontroltool:accesscontroltool-bundle:jar:2.1.1:compile
[INFO] |     +- biz.aQute.bnd:bndlib:jar:2.3.0:compile
[INFO] |     |  \- org.osgi:org.osgi.core:jar:4.3.1:compile`

The access control tool should exclude those dependencies to prevent errors when being used as a dependency.

### LINKED ISSUE #318: ERROR org.apache.felix.metatype fromDocuments: Error accessing document *.properties : XML parsing exception while reading metadata when starting a bundle
Since AEM 6.4 the following error is being emitted whenever the ACTool bundle is being started

```
 *ERROR* [qtp1110251377-63] org.apache.felix.metatype fromDocuments: Error accessing document bundle://547.0:0/OSGI-INF/metatype/biz.netcentric.cq.tools.actool.impl.AcInstallationServiceImpl.properties : XML parsing exception while reading metadata: unterminated entity ref (position:TEXT #Thu May 17 18:0...@2:106 in bundle://547.0:0/OSGI-INF/metatype/biz.netcentric.cq.tools.actool.impl.AcInstallationServiceImpl.properties)
```

The reason for that is a changed handling of the metatype implementation in Apache Felix (https://issues.apache.org/jira/browse/FELIX-5702). Unfortunately the maven-scr-plugin in versions < 1.25 incorrectly generates properties files in the wrong location (https://issues.apache.org/jira/browse/FELIX-5700).

As fix we should either migrate fully to OSGi R6 annotations only (preferably) and get rid of maven-scr-plugin at all or upgrade maven-scr-plugin to at least version 1.25.

kwin:
get rid of maven-scr-plugin, migrate from maven-bundle-plugin to
bnd-maven-plugin, migrate to filevault-package-maven-plugin

Update all Maven plugins to most recent versions.

This closes #318, closes #314

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,2 +1,3 @@
 # standard maven build with travis
 language: java
+jdk: oraclejdk8
--- a/accesscontroltool-bundle/pom.xml
+++ b/accesscontroltool-bundle/pom.xml
@@ -233,6 +233,7 @@ Bundle-Description: ${project.description}
 Bundle-DocURL: https://github.com/Netcentric/accesscontroltool
 Bundle-License: Eclipse Public License 1.0
 Bundle-Vendor: Netcentric
+Bundle-SymbolicName: ${project.groupId}.${project.artifactId}
 # export all versioned packages by default
 -exportcontents: ${packages;VERSIONED}
 Import-Package: org.apache.jackrabbit.oak.spi.security.authentication.external;version="[1.0,3)",*
--- a/accesscontroltool-bundle/pom.xml
+++ b/accesscontroltool-bundle/pom.xml
@@ -3,7 +3,8 @@
     and the accompanying materials are made available under the terms of the
     Eclipse Public License v1.0 which accompanies this distribution, and is available
     at http://www.eclipse.org/legal/epl-v10.html -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd ">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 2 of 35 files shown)


kwin:
Please notice that this is not 100% backwards compatible:
1.  The old configuration PID is no longer supported
2. nrOfSavedHistory and DumpService.nrOfSavedDumps changed from String to int (https://github.com/Netcentric/accesscontroltool/pull/324/files#diff-0d24136154006b3813f897e2077c7339L44 and https://github.com/Netcentric/accesscontroltool/pull/324/files#diff-94d8b01328883d17c67a74b6fd3c7c82L91)

While for the former a fix is feasible it has the drawback that old configuration would basically take precedence of new config (if both exist), for the latter issue a fix would require still type conversion while reading which is nasty. Also we would not correctly identify the type of the metatype as semantically only a numeric value makes sense.
@ghenzler  WDYT?

ghenzler:
I think we should support the old config pid somehow (but log a warning about it)

kwin:
Update: Actually the type conversion should not be an issue, as this is done transparently by SCR (compare with OSGi R6 Comp, $112.8.2.2)

kwin:
Reopened PR for the legacy PID support in https://github.com/Netcentric/accesscontroltool/pull/327

PR #329, opened 2018-10-30
Feature/210 allow to depend on accesscontroltool package with ensured bundle install

### LINKED ISSUE #210: Allow package installation of config package (with yaml files) without the need to have AC Tool package pre-installed
Big fan of your project - love it!

A pain point I'm running into is the installation of ACTool using the content-package-maven-plugin (I want to avoid installing the package via the UI). 

I've specified the following dependency:
```
<dependencies>
    <dependency>
        <groupId>biz.netcentric.cq.tools.accesscontroltool</groupId>
	<artifactId>accesscontroltool-package</artifactId>
	<version>2.0.1</version>
	<type>zip</type>
	<exclusions>
	    <exclusion>
	        <groupId>biz.netcentric.cq.tools.accesscontroltool</groupId>
	        <artifactId>accesscontroltool-bundle</artifactId>
	    </exclusion>
        </exclusions>
    </dependency>
</dependencies>
```
And the following plugin configuration:
```
<plugin>
    <groupId>com.day.jcr.vault</groupId>
    <artifactId>content-package-maven-plugin</artifactId>
    <extensions>true</extensions>
    <configuration>
        <group>mygroup</group>
        <useProxy>false</useProxy>
        <filterSource>src/main/content/META-INF/vault/filter.xml</filterSource>
        <targetURL>http://${sling.host}:${sling.port}/crx/packmgr/service.jsp</targetURL>
       <properties>
            <installhook.actool.class>biz.netcentric.cq.tools.actool.installhook.AcToolInstallHook</installhook.actool.class>
       </properties>
       <subPackages>
           <subPackage>
                <groupId>biz.netcentric.cq.tools.accesscontroltool</groupId>
                <artifactId>accesscontroltool-package</artifactId>
                <filter>true</filter>
           </subPackage>
      </subPackages>
    </configuration>
</plugin>
```
However, the hook executes before the sub package is installed so the groups/ACLs are not applied. 
```
*INFO* [qtp282918978-3407] org.apache.jackrabbit.vault.packaging.impl.InstallHookProcessorImpl Loading Hook actool: Main-Class = biz.netcentric.cq.tools.actool.installhook.AcToolInstallHook
*ERROR* [qtp282918978-3407] org.apache.jackrabbit.vault.packaging.impl.InstallHookProcessorImpl Error while initializing hook: org.apache.jackrabbit.vault.packaging.PackageException: hook's main class biz.netcentric.cq.tools.actool.installhook.AcToolInstallHook not found: actool
```
Have you attempted to install ACTool this way? What would you recommend?

ghenzler:
Closes #210

### DIFF
--- a/accesscontroltool-package/pom.xml
+++ b/accesscontroltool-package/pom.xml
@@ -70,10 +70,11 @@
                         </goals>
                         <configuration>
                             <artifactItems>
+                                <!-- See http://sling.apache.org/documentation/bundles/installer-provider-installhook.html -->
                                 <artifactItem>
                                     <groupId>org.apache.sling</groupId>
                                     <artifactId>org.apache.sling.installer.provider.installhook</artifactId>
-                                    <version>1.0.0-SNAPSHOT</version>
+                                    <version>1.0.0</version>
                                 </artifactItem>
                             </artifactItems>
                             <outputDirectory>${project.build.directory}/vault-work/META-INF/vault/hooks</outputDirectory>
@@ -91,7 +92,8 @@
                     <filterSource>src/main/META-INF/vault/filter.xml</filterSource>
                     <properties>
                         <acHandling>merge_preserve</acHandling>
-                        <installPathRegex>/apps/netcentric/actool/install.*</installPathRegex>
+                        <installPathRegex>/apps/netcentric/actool/.*</installPathRegex>
+                        <waitForOsgiEventsQuietInSec>5</waitForOsgiEventsQuietInSec>
                     </properties>
                     <embeddeds>
                         <embedded>
--- a/accesscontroltool-package/pom.xml
+++ b/accesscontroltool-package/pom.xml
@@ -69,7 +69,9 @@
                     <filterSource>src/main/META-INF/vault/filter.xml</filterSource>
                     <properties>
                         <acHandling>merge_preserve</acHandling>
-                        <installPathRegex>/apps/netcentric/actool/.*</installPathRegex>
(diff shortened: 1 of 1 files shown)


ghenzler:
@kwin Please test this branch with `mvn clean install -PautoInstallPackage -Dvault.verbose`

accesscontroltool-package/pom.xml
```diff
@@ -44,6 +69,8 @@
                     <filterSource>src/main/META-INF/vault/filter.xml</filterSource>
                     <properties>
                         <acHandling>merge_preserve</acHandling>
+                        <installPathRegex>/apps/netcentric/actool/.*</installPathRegex>
```
kwin: Can’t we restrict further? I think only covering the install folder would make sense here

ghenzler: The idea was to have everything installed - however you are right, strictly speaking the two bundles should be enough since for the install hook case, the system user config is not needed (runs as same user as the package manager is called with) and hence it should be fine, will test this.

ghenzler: As it turns out, `AcHistoryServiceImpl.persistHistory()` is currently not using the session from the package manager (but rather the service session). In a way it is nicer to have the path as it is anway: that way after package is installed in package manager the AC Tool is fully installed with all its bundles and configurations (conceptually better).

ghenzler: Made the filter more explicit in ea81c26e8da7369

PR #336, opened 2018-11-19
#335 Make password optional for users

### LINKED ISSUE #335: Make password for users (non-service users) optional
Although most of the time users should not be managed by AC Tool for production (LDAP/SSO integrations are usually in place to take care of that and  only to link to AC Tool-created role-groups), sometimes it can be useful to pre-create production users for those integrations and pre-assign roles to them. 

At the moment password is mandatory for non-system users, this shall be made optional (although the javadoc of [UserManager](https://helpx.adobe.com/experience-manager/6-4/sites/developing/using/reference-materials/javadoc/org/apache/jackrabbit/api/security/user/UserManager.html#createUser-java.lang.String-java.lang.String-java.security.Principal-java.lang.String-) states that password is mandatory, it really can be null and e.g. the LDAP integration of AEM exactly calls it like that which under the hood will just create a user node without `rep:password` property)

ghenzler:
Closes #335


ghenzler:
Updated documentation in bc52f2e

Issue #342, opened 2019-01-07
RUNMODES variable does not work when used in if condition with containsItem


craghav:
The below code snippet doesn't work
- DEF localVal=[default]   
- FOR item IN [prod, nosamplecontent]:
       - IF ${containsItem(RUNMODES, item)}:
          - DEF localVal=[newVal]

otarsko:
@craghav I assume that the problem you have is that `localVal` doesn't change its value to `[newVal]` outside of `FOR` loop. This happens because loops have their own scope, so any changes done within them will not affect variables outside. For example:

```
- DEF localVal=default

- group_config: 
    - FOR item IN [crx3, nosamplecontent]:
        - IF ${containsItem(RUNMODES, item)}:
            - DEF localVal=newval

    - testgroup-${localVal}: 
        - path: /home/groups/test
```

Will result into `testgroup-default` group created. But:

```
- DEF localVal=default

- group_config: 
    - FOR item IN [crx3, nosamplecontent]:
        - IF ${containsItem(RUNMODES, item)}:
            - DEF localVal=newval
            - testgroup-${localVal}: 
                - path: /home/groups/test
```

will create `testgroup-newvals`.

craghav:
But the behaviour of localVal in the below scenario will be updated to the new value when used outside the loop. Should the behaviour not be the same when used with containItems too ?
`- DEF localVal=[default]
- FOR item IN $RUNMODES:
      - IF ${startsWith('prod', item)}:
         - DEF localrunmodes=[newVal]
`

otarsko:
@craghav in later case, you should use `${RUNMODES}` to make it work properly (see example https://github.com/Netcentric/accesscontroltool/blob/develop/accesscontroltool-bundle/src/test/resources/test-loop-over-runmodes.yaml)

AC Tool will not recognise "FOR" loop here, because variable is not used properly, and loop scoping will not be applied.

ghenzler:
@craghav please reopen if the described approach does not work for you

PR #347, opened 2019-01-09
#342: Add information about loop's variable scope. Add documentation about predefined variables.

otarsko:


ghenzler:
thanks for the docu update @otarsko

PR #354, opened 2019-01-21
Avoid to get declared members twices for huge groups

stefanseifert:
Scenario: We have a group with 80000 members (as a starting point, more in the future).
The group itself and all other groups in the system are maintained via ACTool.

problem: when this group is processed, the ACtool processing pauses for several minutes at this groups. looking into the code we detected that those 80000 users are fetched twice - first as part of all declared members, second again to filter out the regular users of the declared members.

this PR optimizes the code path again to filter out the regular users in the first place. i'm not sure if it was by intention that this fetching took place twice, please cross-check.


ghenzler:
Thanks @stefanseifert for the PR, after having a closer look I simplified the code further in 2b2e70e3034aeb Please test branch `develop` if it runs fine for your use case

stefanseifert:
lgtm - thanks!

stefanseifert:
@ghenzler: is it possible to do an new release with this fix? we need it in our production systems ...

ghenzler:
@stefanseifert Created release [2.3.2](https://github.com/Netcentric/accesscontroltool/releases/tag/2.3.2) for this issue

stefanseifert:
thanks a lot!

PR #363, opened 2019-04-16
Add property description for listener service to get prop name

bcsott:
Tried to configure the Upload Listener service. The Service status lacks a description. Therefore you have to look in the source or create and search for the config.


ghenzler:
looks good, thanks @bcsott!

Issue #364, opened 2019-05-14
Allow for commas in group name


gwydi:
In our AEM System we have group names that correspond with their LDAP DN.
This means they contain commas.

I now would like to use the following line in the yaml configuration.

`isMemberOf: cn=Author,ou=...,ou=...,o=...`

Because `isMemberOf` allows comma separated lists the individual parts are treated as individual group.

I tried to use single `'` and double `"` quotes as well as backslashes `\` to escape the commas without success. 

Is there a way to get around this issue?

gwydi:
I have implemented a way to escape the commas. I have no permission to push my branch though.

kwin:
@gwydi Feel free to open a PR as outlined in https://help.github.com/en/articles/creating-a-pull-request-from-a-fork. You need to create the branch in your fork.

kwin:
I am pretty sure though that something like 
```
isMemberOf: 'cn=Author,ou=...,ou=...,o=...'
```
should work though. According to https://bitbucket.org/asomov/snakeyaml/wiki/Documentation#markdown-header-scalars this is also fully supported by the underlying yaml parser used in AC Tool. For further information please refer to https://yaml.org/spec/1.2/spec.html#id2786942 and particularly to https://yaml.org/spec/1.2/spec.html#id2788859 for why your example cannot work:
> In addition, inside flow collections, or when used as implicit keys, plain scalars must not contain the “[”, “]”, “{”, “}” and “,” characters. These characters would cause ambiguity with flow collection structures.

ghenzler:
@gwydi I think you are right, there is no way you can solve this currently because on YAML-level the `isMemberOf` is just a string that is then split up by `,` by the AC Tool (this is for historic reasons).

IMHO the best fix would be to allow both a list of strings and a simple string at YAML level for `isMemberOf` - if you choose a string you cannot use `,` in authorizable names, but if you use a YAML list you then could...

gwydi:
@ghenzler I implemented a fix that ignores `\,` when splitting the string. 
But i do agree, fixing the issue on YAML level would be better.

I could try to look into that.

gwydi:
I created a PR from my fork.
This fix solves our problem with the commas and we will use it until this repo has a fix for the issue.
Feel free to reject the PR if you are going to implement a fix in another way.

best regards
Marcel

ghenzler:
[...]

PR #366, opened 2019-05-23
#364 Ignoring escaped commas in isMemberOf and memberOf 

gwydi:
Fixing behavior described in #364.


ghenzler:
@gwydi Sorry for only coming back to you now - so this PR contains a lot of whitespace changes (minor thing) but more importantly I think we should not introduce another workaround, rather we should allow for proper yaml lists in parallel to the string (see also https://github.com/Netcentric/accesscontroltool/issues/364#issuecomment-523483836)

ghenzler:
Opened now #390 as alternative

Issue #370, opened 2019-08-20
Ignore trailing slashes for same path definitions


luckyluke80:
The following configuration ignores the permissions for groupB and groupC and possibly all following group definitions with the same path.

```
- def groupPath="/home/groups/slashissue/fragments"

- group_config:

    - groupA:
      - path: ${groupPath}
  
    - groupB:
      - path: ${groupPath}

    - groupC:
      - path: ${groupPath}

- ace_config:

    - groupA:
        - path: /content/dam/we-retail/
          permission: allow
          actions: read

    - groupB:
        - path: /content/dam/we-retail
          permission: allow
          actions: read,modify,create,delete

    - groupC:
        - path: /content/dam/we-retail
          permission: allow
          actions: acl_read,replicate
```

It's currently not allowed to mix trailing slashes for the same path. The permissions are applied if no trailing slashes for the same path are defined or if all same path's have trailing slashes.

ghenzler:
The easiest would be to just normalise the input at https://github.com/Netcentric/accesscontroltool/blob/cbc4d74e180ee775deefeeb81c1fe6bb9f07f4e9/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/configreader/YamlConfigReader.java#L303 (and chop off a trailing slash if present as conceptually, it is meant to be treated in the exact same way)

PR #371, opened 2019-08-22
370: removing trailing slashes while reading the configuration

luckyluke80:
the Pull Request contains the following changes:

- the proposal to cleanup the trailing slashes when reading the config
- an extended unit test, which will ensure, that all path's are the same
- an extended yaml file

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/configreader/YamlConfigReader.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/configreader/YamlConfigReader.java
@@ -300,9 +300,7 @@ public class YamlConfigReader implements ConfigReader {
                                                   // AcConfiguration.ensureAceBeansHaveCorrectPrincipalNameSet()
 
         String jcrPath = getMapValueAsString(currentAceDefinition, ACE_CONFIG_PROPERTY_PATH);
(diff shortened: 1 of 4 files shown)


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/configreader/YamlConfigReader.java
```diff
@@ -293,31 +293,26 @@ protected AuthorizableConfigBean getNewAuthorizableConfigBean() {
...
 
         aclBean.setAuthorizableId(authorizableId);
         aclBean.setPrincipalName(authorizableId); // to ensure it is set, later corrected if necessary in
                                                   // AcConfiguration.ensureAceBeansHaveCorrectPrincipalNameSet()
 
-        aclBean.setJcrPath(getMapValueAsString(currentAceDefinition,
-                ACE_CONFIG_PROPERTY_PATH));
+        String jcrPath = getMapValueAsString(currentAceDefinition, ACE_CONFIG_PROPERTY_PATH);
+        String cleanJcrPath =  StringUtils.removeEnd(jcrPath, "/");
+        LOG.debug("cleaned jcr path: {} to {}", jcrPath, cleanJcrPath);
```
ghenzler: @luckyluke80 this log is misleading (it will also show if nothing was removed) - I think it's best to just remove the log

ghenzler:
Closes #370

PR #374, opened 2019-10-11
optionally execute installhook at phase "INSTALLED"

### LINKED ISSUE #287: InstallHook should be executed in  phase "INSTALLED" rather than "PREPARE"
Currently the install hook for applying the yaml files is being executed in Phase "Prepare" (https://jackrabbit.apache.org/filevault/apidocs/org/apache/jackrabbit/vault/packaging/InstallContext.Phase.html). The reason for that is most probably due to the description in https://jackrabbit.apache.org/filevault/apidocs/org/apache/jackrabbit/vault/packaging/InstallHook.html  which states that only from Phase "Prepare" you can throw exceptions which lead to package installation failures.

The drawback of using PREPARE though is that 

1. you cannot create/update some initial content through a regular package structure and afterwards set ACLs for those in the same package. 
1. You cannot use repository structures as workaround for missing support for global variables (#257)
1. Initial content should be defined as regular content to be able to depend on it for application package type (https://issues.apache.org/jira/browse/JCRVLT-170). Those package types are not allowed to contain uncovered roots!

We should check again if PREPARE is really the only phase where Exception are being properly propagated.

kwin:
upgrade bnd plugin to version 4.3.0
This closes #287


accesscontroltool-bundle/pom.xml
```diff
@@ -287,6 +285,11 @@ Import-Package: org.apache.jackrabbit.oak.spi.security.authentication.external;r
                         <goals>
                             <goal>baseline</goal>
                         </goals>
+                        <configuration>
+                            <fullReport>true</fullReport>
+                            <!-- TODO: skip baselining prior to release to disable bundle check, see https://github.com/bndtools/bnd/issues/3446 -->
+                            <skip>false</skip>
+                        </configuration>
```
ghenzler: Why do we have to touch the bnd tooling for this feature?

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/installhook/impl/OsgiAwareInstallHook.java
```diff
@@ -45,22 +47,45 @@ public OsgiAwareInstallHook() throws ClassCastException {
 		}
 	}
 
+	// TODO: make protected
```
ghenzler: Can we not just make those changes now if they are needed? (or let's don't add todos but rather just create an issue `Make OsgiAwareInstallHook more restrictive`)

ghenzler (changes requested):

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/installhook/impl/OsgiAwareInstallHook.java
kwin: This requires a major version increase, I don’t want to do right now. I will create an issue though

accesscontroltool-bundle/pom.xml
kwin: I can separate those changes into a dedicated PR. Any general concerns with the  update?

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/installhook/impl/OsgiAwareInstallHook.java
kwin: Done in https://github.com/Netcentric/accesscontroltool/issues/394.

ghenzler:
Fixed a few minor things (2d068bd795 e.g. version in docu and removed todos) and merged develop into this branch for testing - looks good to me

PR #376, opened 2019-10-11
Fix tag paths in examples to use

gayathrigs27:
Changed the value in the config file


ghenzler:
@gayathrigs27 used the more complete pr #377

Issue #375, opened 2019-10-11
Fix tag paths in examples to use `/content/cq:tags` 


ghenzler:
Fix tag paths in examples to use `/content/cq:tags` instead of old `/etc/tags`

e.g. https://github.com/Netcentric/accesscontroltool/blob/develop/accesscontroltool-exampleconfig-package/src/main/jcr_root/apps/netcentric/actool-exampleconfig/realProject/global.author/config.yaml#L37

gayathrigs27:
raised a pr for it

PR #377, opened 2019-10-11
Fix tag paths in examples to use '/content/cq:tags'

AmyrAhmady:
Title says it all
related issue: #375


ghenzler:
Thanks @AmyrAhmady for your contribution

PR #392, opened 2019-11-04
Avoid the requirement for the additional index rep:ACL

### LINKED ISSUE #386: Avoid the requirement for the additional index `rep:ACL`
The index `rep:ACL` in package `accesscontroltool-oakindex-package` currently works well, however for future AEM deployment scenarios, it would be great to be able to do without an additional index.

ghenzler:
Closes #386

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/QueryHelper.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/QueryHelper.java
@@ -10,7 +10,6 @@ package biz.netcentric.cq.tools.actool.helper;
 
 import static biz.netcentric.cq.tools.actool.history.PersistableInstallationLogger.msHumanReadable;
 
-import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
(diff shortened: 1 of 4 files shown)


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/QueryHelper.java
```diff
@@ -73,32 +77,47 @@
...
+            if (session.nodeExists(HOME_REP_POLICY)) {
+                nodes.add(session.getNode(HOME_REP_POLICY));
             }
 
+            boolean indexForRepACLExists = session.nodeExists(OAK_INDEX_PATH_REP_ACL);
+            LOG.debug("Index for repACL exists: {}",indexForRepACLExists);
+            String queryForAClNodes = indexForRepACLExists ? 
+                    "SELECT * FROM [rep:ACL] WHERE ISDESCENDANTNODE([%s])" : 
+                    "SELECT ace.* FROM [rep:ACE] AS ace WHERE ace.[rep:principalName] IS NOT NULL AND ISDESCENDANTNODE(ace, [%s])";
+            LOG.debug("Query to obtain all ACLs: {}", queryForAClNodes);
```
ghenzler, line 103: For now we leave the old query for the case  the index is there (to reduce the impact for existing users if there are problems)

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/QueryHelper.java
```diff
@@ -73,32 +77,47 @@
...
+                    continue;
+                }
+                
+                String query = String.format(queryForAClNodes, path);
+
+                long startTime1 = System.currentTimeMillis();
+                Set<Node> nodesResult = indexForRepACLExists ? 
+                        getNodes(session, query, Query.JCR_SQL2):
+                        getDistinctParentNodes(session, query, Query.JCR_SQL2);
+                LOG.debug("Query to find ACLs under {} ran in {}ms (count ACLs: {})", path, System.currentTimeMillis()-startTime1, nodesResult.size());
```
ghenzler, line 116: Performance tests with an instance with 2 Mio pages and a fair amount of ACLs in them looked ok (still under 1sec with this approach, >100ms with the old approach)

ghenzler: the test above was with a publish instance (and 2 Mio pages), it turns out that on an author instance with the same content the performance can degrade to around 8 seconds. I think this is still acceptable (also because for the case the index is there, it works with the old speed, see https://github.com/Netcentric/accesscontroltool/pull/392#discussion_r341921557)

PR #393, opened 2019-11-08
Support for custom Oak privileges

YegorKozlov:
This is an exotic case, but it came from a real project. 
AC Tool can set custom privileges just fine, the only twist I had to do was to create them manually. Initially it was a script to run in AEM Fiddle. Something like :
```
<%@include file="/libs/foundation/global.jsp"%><%
%><%@page session="false" contentType="text/html; charset=utf-8" 
	pageEncoding="UTF-8"
    import="org.apache.sling.api.resource.*,
    java.util.*,
    javax.jcr.*,
    org.apache.jackrabbit.api.security.authorization.PrivilegeManager,
    org.apache.jackrabbit.api.*"%><pre><%

    String[] privileges = {"sling:feature1", "sling:feature2"};

    Session session = resourceResolver.adaptTo(Session.class);
    PrivilegeManager privilegeManager = ((JackrabbitWorkspace)session.getWorkspace()).getPrivilegeManager();
    for(String privilegeName : privileges){
      privilegeManager.registerPrivilege(privilegeName, false, new String[0]);
      out.println("created " + privilegeName);
    }

%></pre>
```

Ideally, custom privileges and ACEs should be all set in Yaml and installed in one go and this patch introduces support for this.

```
- privilege_config:

    - sling:feature:

    - sling:anotherFeature:
        - isAbstract: false
          declaredAggregateNames: jcr:read,jcr:write


- group_config:

    - sling-feature-users:
        - name:

- ace_config:

    - sling-feature-users:

        - path: /content
          permission: allow
          privileges: sling:feature

        - path: /content/dam
          permission: allow
          privileges: sling:anotherFeature
```
There is a new section `privilege_config` which is evaluated before installation of authorazibles. A privilege is defined by a (name, isAbstract, aggregateNames) tripplet and the impementation passes these arguments to Oak's PrivilegeManager#registerPrivilege
The built-in and created privileges live under /jcr:system/rep:privileges and you can check the created nodes in CRXDE.

Creating a privilege is a one-way operation. PrivilegeManager does not provide a "removePrivilege" operation, so from AC Tool prospective this cannot be purged.

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/PrivilegeInstaller.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/PrivilegeInstaller.java
@@ -1,0 +1,21 @@
+/*
+ * (C) Copyright 2015 Netcentric AG.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package biz.netcentric.cq.tools.actool.authorizableinstaller;
+
+import biz.netcentric.cq.tools.actool.configmodel.PrivilegeConfig;
+import biz.netcentric.cq.tools.actool.history.InstallationLogger;
+
+import javax.jcr.RepositoryException;
+import javax.jcr.Session;
+
+public interface PrivilegeInstaller {
+
+    void installPrivileges(PrivilegeConfig config, Session session, InstallationLogger installLog)
+            throws RepositoryException;
+}
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/PrivilegeInstallerImpl.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/PrivilegeInstallerImpl.java
@@ -1,0 +1,51 @@
+/*
+ * (C) Copyright 2015 Netcentric AG.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ */
+package biz.netcentric.cq.tools.actool.authorizableinstaller.impl;
+
+import biz.netcentric.cq.tools.actool.authorizableinstaller.PrivilegeInstaller;
+import biz.netcentric.cq.tools.actool.configmodel.PrivilegeBean;
+import biz.netcentric.cq.tools.actool.configmodel.PrivilegeConfig;
+import biz.netcentric.cq.tools.actool.history.InstallationLogger;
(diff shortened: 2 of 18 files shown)


kwin:
Please rather use the feature of FileVault Packages for registering privileges: https://issues.apache.org/jira/plugins/servlet/mobile#issue/JCRVLT-384

YegorKozlov:
FileVault and AC Tool are not mutually exclusive.

I can use FileVault for registering privileges, but it still would be a separate step, i.e. *something* to define outside of the Yaml files.
Since the AC Tool provides a friendly Yaml interface for managing authorizables and permissions, it might be a good idea to support registering custom privileges directly in Yaml. 

Also, privilege.xml is well hidden gem in FileVault :) I've been using it for years and didn't know it was possible.  The only example I could find was https://github.com/apache/jackrabbit-filevault/blob/trunk/vault-core/src/test/resources/org/apache/jackrabbit/vault/packaging/integration/testpackages/privileges.zip

ghenzler:
Hi @YegorKozlov, in a way it makes sense to have it all in one place, on the other hand if it is possible to easily register privileges with a package, a simple package with the config package having a dependency to it would do. 

From what I can tell when using /crx/explorer/config/privileges.jsp or looking at the javadoc [1], it looks like privileges can only be registered (but not unregistered/cleaned up). 

I tried to generate a package that contains a `privileges.xml` file with the package mgr but it seems like it never generates such a file (created content with a manually registered privilege and create a package with that, but no `privileges.xml` file was in the package, just the privilege itself was referenced. @kwin Would you have an example file?

[1] https://helpx.adobe.com/experience-manager/6-5/sites/developing/using/reference-materials/javadoc/org/apache/jackrabbit/api/security/authorization/PrivilegeManager.html

YegorKozlov:
Hi @ghenzler 

I was able to get privileges.xml working, but I had to craft the package manually. Packmgr does not support it, the only option is to define the file manually and put in your maven project next to filter.xml. 

You are right that installing privileges is a one-way operation. I thought about extending it and adding a PrivielegeManager#removePrivilege operation, but it's not that simple. The interface belongs to Jackrabbit, the impl is by Oak, so you cannot just add a new operation in the interface. 

I'm fine to revoke my patch and use privileges.xml, but the site should document this feature. In most cases users deploy the Yaml scripts from a package which can also auto-install privileges.  A bonus from privileges.xml is that can install namespaces too, e.g. for  `netcentric:coolfeature` it will ensure both the `netcentric `namespace and the `netcentric:coolfeature`  privilege.

Regards,
Yegor

kwin:
@YegorKozlov Feel free to come up with a documentation PR for that.

YegorKozlov:
> @YegorKozlov Feel free to come up with a documentation PR for that.

Sure @kwin  I'll see what I can do.

kwin:
Privileges.xml files within packages are now documented at <http://jackrabbit.apache.org/filevault/privileges.html>.

ghenzler:
I think we have consensus that it's better to work with the privileges.xml file directly (instead of introducing a custom new syntax for this in the AC Tool). One of the main advantages of maintaining ACLs & authorizables with the AC tool is that it can also remove (=cleanup) old entries, for privileges this advantage generally does not exist as it's not possible to remove privileges from a repository. 

@YegorKozlov if you like to come up with a PR for the documentation on how to do this best, this would be great! :)

PR #399, opened 2019-12-02
Split up package into application, content and container part

### LINKED ISSUE #398: Split up packages by package type (separate content from apps)
See https://issues.apache.org/jira/browse/JCRVLT-170 for details

kwin:
This closes #398


accesscontroltool-package/pom.xml
```diff
@@ -24,15 +24,27 @@
 
     <artifactId>accesscontroltool-package</artifactId>
     <packaging>content-package</packaging>
-    <name>Access Control Tool Package</name>
-    <description>Core package of the AC Tool containing the bundle, system user and ACLs for it.</description>
+    <name>Access Control Tool Package - Container</name>
```
ghenzler: Looks all good except that I would leave the name here as is (and not add ` - Container`), this is also the package name in package mgr and it might confuse people if it changes (as really nothing changes from a user perspective)

ghenzler (changes requested):

Issue #394, opened 2019-11-12
Clean up API


kwin:
Currently exported interfaces rely on classes which are not being exported and therefore lead to an inconsistent API. One example is `biz.netcentric.cq.tools.actool.installhook.impl.OsgiAwareInstallHook.java` which is parent of the exported `AcToolInstallHook` and exposes some methods (i.e. methods which are not private).

PR #406, opened 2019-12-18
#405 Allow to apply the AC tool config upon startup

### LINKED ISSUE #405: Support running the AC Tool automatically upon server start
When using the composite node store [1] with immutable containers (and immutable /apps and /libs paths), we need a new trigger to apply the AC tool configuration. As there is no regular deployment anymore (as we have today when installing yaml files with a package via package manager) the startup of the instance is the only trigger to ensure the AC configuration is applied correctly. 

To create this trigger a new bundle `accesscontroltool-startuphook-bundle` has to be created that that triggers the AC tool installation after yaml files and AC tool itself are available (at OSGi start level > 20)

* When preparing a composite node store (during build of image), the startup hook can modify all paths (including /apps and /libs, it's not a composite node store yet)
* When starting with a composite node store the startup hook must run again (to ensure groups/users and correct ACLS are created again for non apps/libs-parts). To ensure the AC tool does not run upon every startup (although the run is incremental and would not make actual changes this would impose significant unnecessary load during auto-scaling), a hash over all config files shall be created and the AC tool only be applied if there are actual changes to the configuration.

[1] https://jackrabbit.apache.org/oak/docs/nodestore/compositens.html

ghenzler:
Closes #405 #407 #408


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/runtime/RuntimeHelper.java
```diff
@@ -0,0 +1,76 @@
...
+public class RuntimeHelper {
+    
+    private static final String RUNMODE_CLOUD_READY = "cloud-ready";
+
+    public static boolean isCompositeNodeStore(Session session) {
+        
+        try {
+            Node appsNode = session.getNode("/apps");
+            // see https://issues.apache.org/jira/browse/OAK-6563
+            boolean isCompositeNode = !session.hasCapability("addNode", appsNode, new Object[] { "nt:folder" });
```
kwin: This could also mean that the given session just does not have the according rights, isn't it? That should be explicitly mentioned or a check for the according rights must be added here as well. 
For simplicity reasons I would always rely on the service user internally and not pass the session as argument....

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/runtime/RuntimeHelper.java
```diff
@@ -0,0 +1,76 @@
...
+        return bundleContext.getBundle(Constants.SYSTEM_BUNDLE_ID).adapt(FrameworkStartLevel.class).getStartLevel();
+    }
+    
+    public static boolean isCloudReadyInstance(SlingSettingsService settingsService) {
+        Set<String> runModes = settingsService.getRunModes();
+        boolean isCloudReady = runModes.contains(RUNMODE_CLOUD_READY);
+        return isCloudReady;
+    }
+    
+    public static Node copyNode(Node src, Node dstParent) throws RepositoryException {
```
kwin: Can't you use https://docs.adobe.com/docs/en/spec/jsr170/javadocs/jcr-2.0/javax/jcr/Workspace.html#copy(java.lang.String,%20java.lang.String)

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/runtime/RuntimeHelper.java
```diff
@@ -0,0 +1,76 @@
...
+        for (NodeIterator iter = src.getNodes(); iter.hasNext();) {
+            Node n = iter.nextNode();
+            if (!n.getDefinition().isProtected()) {
+                copyNode(n, dst);
+            }
+        }
+        return dst;
+    }
+
+    public static Property copyProperty(Property src, Node dstParent) throws RepositoryException {
```
kwin: IMHO not necessary

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/impl/AcConfigChangeTracker.java
```diff
@@ -0,0 +1,81 @@
...
+    private String createHashOverConfigFiles(Map<String, String> configFiles) throws Exception {
+        StringBuilder buf = new StringBuilder();
+        for (Map.Entry<String,String> test : configFiles.entrySet()) {
+            buf.append(test.getKey()+"\n"+test.getValue());
+        }
+
+        return md5(buf.toString());
+    }
+    
+    public String md5(String input) throws Exception {
```
kwin: prrivate?

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/impl/AcInstallationServiceImpl.java
```diff
@@ -126,16 +130,15 @@
             description="Service that installs groups & ACEs according to textual configuration files",
             id = CONFIG_PID)
     protected static @interface Configuration {
-        @AttributeDefinition(name="Configuration storage path", description="CRX path where ACE configuration gets stored")
+        @AttributeDefinition(name="Configuration path", description="JCR path where the AC Tool YAML configuration files are stored")
```
kwin: We should mention "root path" and also add some clarification around traversal.

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/impl/AcInstallationServiceImpl.java
```diff
@@ -207,22 +222,38 @@ public InstallationLog apply(String configurationRootPath, String[] restrictedTo
         return installLog;
     }
 
-    /** Common entry point for JMX and install hook, 
-     * TODO: should not be exported as using non-API class PersistableInstllationLogger */
```
kwin, line 211: This TODO should be kept!

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/jmx/AceServiceMBean.java
```diff
@@ -45,8 +45,12 @@ String applyRestrictedToPaths(
     String applyRestrictedToPaths(@Name("configurationRootPath") @Description("The configuration root path") String configurationRootPath,
             @Name("paths") @Description("comma separated list of paths to apply the configuration to, other paths will be skipped") String restrictedToPaths);
 
-    @Description("Purges the AccessControlList of the given path, if existing")
+    @Description("Applies the ACE configuration if it has not changed")
```
kwin: "if it has not  changed"?

accesscontroltool-install-packages/accesscontroltool-cr-package/pom.xml
```diff
@@ -0,0 +1,98 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kwin: please don't use abbreviations no one will understand. Why not accesscontroltool-cloud-package? The official name is now AEM in the cloud!

kwin (changes requested):

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/impl/AcInstallationServiceImpl.java
ghenzler: @kwin please add an issue here (TODOs get forgotten, issues are better in that respect)

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/impl/AcConfigChangeTracker.java
ghenzler: will fix this

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/runtime/RuntimeHelper.java
ghenzler: > For simplicity reasons I would always rely on the service user internally and not pass the session as argument....

The session can be either from the install hook (packmgr) or the service session  - I really want to make sure that the exact session that is also used for installation is used here

> This could also mean that the given session just does not have the according rights, isn't it? 

Correct, will try a combination with `session.hasPermission()` to ensure it's 100% correct

accesscontroltool-install-packages/accesscontroltool-cr-package/pom.xml
ghenzler: `cr` stands for cloud-ready, but yes, `accesscontroltool-cloud-package` is probably better to understand, will adjust it.

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/impl/AcInstallationServiceImpl.java
kwin: There is already an issue: https://github.com/Netcentric/accesscontroltool/issues/394.
Todo's don't hurt though because they have a better context!

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/runtime/RuntimeHelper.java
ghenzler: Improved implementation: https://github.com/Netcentric/accesscontroltool/blob/6db01125d0c038cacc2df27b908ad709b129e026/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/runtime/RuntimeHelper.java#L41

Issue #325, opened 2018-10-22
Tooling: Enforce best practices via osgicheck-maven-plugin


kwin:
The recently released plugin https://github.com/apache/felix/tree/trunk/tools/osgicheck-maven-plugin enforces some best practices with regard to OSGi. It should be included in the build to emit helpful warning and also breaking the build in case of heavily violating best practices.

PR #409, opened 2019-12-31
#402 Touch UI Interface

### LINKED ISSUE #402: Create user interface in Touch UI
The UI available at /system/console/actool (triggering, downloading the dump, listing/showing the dumps) should be made available in Touch UI (to have it for cases where /system/console is not available)

ghenzler:
Closes #402


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/AcToolTouchUiServlet.java
```diff
@@ -0,0 +1,103 @@
...
+
+/** Touch UI Servlet */
+@Component(service = { Servlet.class })
+@SlingServletResourceTypes(resourceTypes = "/apps/netcentric/actool/components/overview", methods = { "GET", "POST" })
+@SuppressWarnings("serial")
+public class AcToolTouchUiServlet extends SlingAllMethodsServlet {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AcToolTouchUiServlet.class);
+
+    @Reference(policyOption = ReferencePolicyOption.GREEDY)
```
kwin, line 40: for static unary references greedy is not necessary!

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/AcToolTouchUiServlet.java
```diff
@@ -0,0 +1,103 @@
...
+    private WebConsoleConfigTracker webConsoleConfigTracker;
+
+    @Reference(policyOption = ReferencePolicyOption.GREEDY)
```
kwin, line 43: for static unary references greedy is not necessary!

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/AcToolTouchUiServlet.java
```diff
@@ -0,0 +1,103 @@
...
+    private AcToolUiService acToolUiService;
+
+    @Override
+    protected void doGet(SlingHttpServletRequest req, SlingHttpServletResponse resp) throws ServletException, IOException {
+
+        if (StringUtils.isBlank(req.getRequestPathInfo().getSuffix())) {
```
kwin, line 49: what is delivered here and why is the suffix itself not evaluated?

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/AcToolTouchUiServlet.java
```diff
@@ -0,0 +1,103 @@
...
+        if (!mayApplyConfig(req)) {
+            resp.sendError(HttpServletResponse.SC_FORBIDDEN, "You do not have sufficent permissions to apply the configuration");
+            return;
+        }
+
+        acToolUiService.doPost(req, resp);
+        LOG.debug("Applied AC tool config via Touch UI by user {}", req.getUserPrincipal());
+    }
+
+    public boolean mayApplyConfig(SlingHttpServletRequest request) {
```
kwin: rather use User as method argument

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/AcToolTouchUiServlet.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @Override
+    protected void doPost(SlingHttpServletRequest req, SlingHttpServletResponse resp) throws ServletException, IOException {
+
+        if (!mayApplyConfig(req)) {
+            resp.sendError(HttpServletResponse.SC_FORBIDDEN, "You do not have sufficent permissions to apply the configuration");
+            return;
+        }
+
+        acToolUiService.doPost(req, resp);
```
kwin, line 67: either use service user or current user if not in according group

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/AcToolUiService.java
```diff
@@ -0,0 +1,273 @@
...
+import org.slf4j.LoggerFactory;
+
+import biz.netcentric.cq.tools.actool.api.InstallationLog;
+import biz.netcentric.cq.tools.actool.dumpservice.ConfigDumpService;
+import biz.netcentric.cq.tools.actool.history.AcHistoryService;
+import biz.netcentric.cq.tools.actool.history.AcToolExecution;
+import biz.netcentric.cq.tools.actool.history.PersistableInstallationLogger;
+import biz.netcentric.cq.tools.actool.impl.AcInstallationServiceInternal;
+
+@Component(service = { AcToolUiService.class })
```
kwin, line 35: I would use inheritance here and use the same base class for web console plugin and touch UI console servlet. For a dedicated service API this service too much depends on HTML and HTTP

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/AcToolWebconsolePlugin.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public static final String TITLE = "AC Tool";
+    public static final String LABEL = AcToolUiService.PAGE_NAME;
+    public static final String CATEGORY = "Main";
+
+    @Reference(policyOption = ReferencePolicyOption.GREEDY)
+    private AcToolUiService acToolUiService;
+
+    @Override
+    protected void doGet(final HttpServletRequest req, final HttpServletResponse resp) throws ServletException, IOException {
```
kwin, line 41: use a method getResource() for static resources (https://github.com/apache/sling-org-apache-sling-installer-console/pull/1/files#diff-25dc1005e61f822bf34b60973099a856R90). This is called by the Felix Web Console implementation automatically!

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/HtmlWriter.java
```diff
@@ -0,0 +1,80 @@
...
+    final PrintWriter pw;
+    final boolean isTouchUi;
+
+    HtmlWriter(final PrintWriter pw, boolean isTouchUi) {
+        this.pw = pw;
+        this.isTouchUi = isTouchUi;
+    }
+
+    void openTable(String id) {
+        pw.println("<table id='"+id+"' " + (isTouchUi ? " is='coral-table'" : " class='content' cellpadding='0' cellspacing='0' width='100%'") + ">");
```
kwin, line 19: rather use fluent api via `append` instead of all the concatenation. Easier to read and should be faster

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/RequestParameters.java
```diff
@@ -0,0 +1,60 @@
...
+import javax.servlet.http.HttpServletRequest;
+
+import org.apache.commons.lang3.StringUtils;
+
+import biz.netcentric.cq.tools.actool.api.AcInstallationService;
+import biz.netcentric.cq.tools.actool.impl.AcInstallationServiceImpl;
+
+public class RequestParameters {
+
+    static RequestParameters fromRequest(HttpServletRequest req, AcInstallationService acInstallationService) {
```
kwin: I would move this to AcToolUiService (i.e. the abstract base class for both Web Console and Touch UI Console) and use dedicated getters.

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/WebConsoleConfigTracker.java
```diff
@@ -0,0 +1,88 @@
...
+    private static final String CONSOLE_ROOT_PROP = "manager.root";
+    private static final String CONSOLE_ROOT_DEFAULT = "/system/console"; 
+    
+    private static final String CONSOLE_SEC_PROVIDER_PID = "org.apache.sling.extensions.webconsolesecurityprovider.internal.SlingWebConsoleSecurityProvider";
+    private static final String CONSOLE_SEC_PROVIDER_USERS_PROP = "users";
+    private static final String CONSOLE_SEC_PROVIDER_GROUPS_PROP = "groups";
+
+    private static final String[] RELEVANT_PIDS = new String[] {CONSOLE_PID, CONSOLE_SEC_PROVIDER_PID};
+
+    @Reference(policyOption = ReferencePolicyOption.GREEDY)
```
kwin, line 29: static unary -> must not be greedy

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/WebConsoleConfigTracker.java
```diff
@@ -0,0 +1,88 @@
...
+    @Override
+    public void configurationEvent(ConfigurationEvent event) {
+        String pid = event.getPid();
+        if(ArrayUtils.contains(RELEVANT_PIDS, pid)) {
+            updateConfig();
+        }
+    }
+
+    public String getWebConsoleRoot() {
+        return webConsoleRoot;
```
kwin: I think it is more complicated to return the real path (https://github.com/apache/felix/blob/c2eedf660e7814824c31558f0e3ea8f4082aa720/webconsole/src/main/java/org/apache/felix/webconsole/internal/servlet/OsgiManager.java#L1086). I would rather not expose the link in the Granite UI console. Within the Felix WebConsole it is available as request attribute with name https://github.com/apache/felix/blob/c2eedf660e7814824c31558f0e3ea8f4082aa720/webconsole/src/main/java/org/apache/felix/webconsole/WebConsoleConstants.java#L142.

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/AcToolUiService.java
```diff
@@ -0,0 +1,273 @@
...
+        }
+    }
+
+    protected void doPost(final HttpServletRequest req, final HttpServletResponse resp) throws IOException, ServletException {
+
+        RequestParameters reqParams = RequestParameters.fromRequest(req, acInstallationService);
+        LOG.info("Received POST request to apply AC Tool config with configurationRootPath={} basePaths={}",
+                reqParams.configurationRootPath, reqParams.basePaths);
+
+        InstallationLog log = acInstallationService.apply(reqParams.configurationRootPath, reqParams.getBasePathsArr(),
```
kwin, line 78: Can't we explicitly pass a resourceResolver/usersession for apply?

kwin (changes requested):

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/AcToolTouchUiServlet.java
ghenzler: having the suffix in touch UI makes the path end with `actool` for both system console and touch UI - this simplifies link handling

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/AcToolTouchUiServlet.java
ghenzler: I have to check this - `osgicheck-maven-plugin` complained about this at some point

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/AcToolUiService.java
ghenzler: see https://github.com/Netcentric/accesscontroltool/issues/402#issuecomment-570176224

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/AcToolTouchUiServlet.java
ghenzler: It's still complaining (and we have the extra greedy in many other places introduced with e.g. a3e446f142209cdfed2 and #325) - so I would leave it in

kwin: I don't understand this, do we really want this service to restart in case a new implementation with a higher ranking is added?

kwin: Ok, it is indeed documented here: https://github.com/apache/felix/tree/trunk/tools/osgicheck-maven-plugin#available-checks. Still I am not convinced for services which are not supposed to have multiple implementation that this should be added, maybe something to discuss on the Felix Mailing List.

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/WebConsoleConfigTracker.java
ghenzler: Makes sense, will change to uses the request attribute instead.

PR #411, opened 2020-01-02
#410 Allow for multiple configuration roots

### LINKED ISSUE #410: Allow to configure multiple configuration root paths
Currently pid `biz.netcentric.cq.tools.actool.impl.AcInstallationServiceImpl` allows to configure exactly one root path in property `AceService.configurationPath`.

For platform setups it is often necessary to use one config set of yaml files per project on the platform. Currently this can be done by either just using the install hook (because that will ignore the config for the root path and just take the yaml files in the package that has the install hook configured) or by running multiple apply runs via API/JMX explicitly providing the root config path.

With the new feature to apply the config upon server start (#405), multiple root paths shall become configurable using the config property `configurationRootPaths` (if only one is needed, obviously a list of one can be configured). When calling `apply()` without explicit config path, a separate run for each path shall be triggered.

ghenzler:
Closes #410


kwin:
I would prefer a configuration factory! That way the roots can be configured in separate configs.

ghenzler:
> I would prefer a configuration factory! That way the roots can be configured in separate configs.

could be done that way (with more refactoring)... the biggest benefit would be to allow for other config values per root path (and not just globally for all root paths)... however we conceptually have the global config in the yaml files themselves, so there are no plans to add OSGi config values (and the one  property `intermediateSaves` that we currently have can be removed with version 3.0.0 since it's not necessary anymore).

kwin (approved):

kwin:
Can you update https://github.com/Netcentric/accesscontroltool/blob/develop/docs/ApplyConfig.md#upload-listener-service as well?

Issue #412, opened 2020-01-02
Remove jdk 8 imports


tomasalmeida:
Trying to compile using jdk 7 gives the following error:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:compile (default-compile) on project accesscontroltool-bundle: Compilation failure: Compilation failure:
[ERROR] /Users/tomas/projects/accesscontroltool/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/impl/AcInstallationServiceImpl.java:[25,26] package java.util.function does not exist
[ERROR] /Users/tomas/projects/accesscontroltool/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/impl/AcInstallationServiceImpl.java:[26,24] package java.util.stream does not exist
```

These classes are not used, so they can be safely removed

PR #413, opened 2020-01-02
#412 remove unused imports

tomasalmeida:
These classes are not used and break code compatibility with jdk7


kwin:
This was introduced by https://github.com/tomasalmeida/accesscontroltool/commit/a1c0c14a044bf005cfc7ed750ae5213a3b97a131. Let us make sure that this does not happen again by leveraging animal sniffer.

ghenzler:
Thanks @tomasalmeida for fixing this (I had also noticed in the meantime, they are also removed in https://github.com/Netcentric/accesscontroltool/pull/411/files, but better twice then never!)

Issue #400, opened 2019-12-03
Upgrade to Java 8 (drop support of AEM 6.1)


kwin:
To leverage classes like [java.util.Base64](https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html) we should compile for Java 8. Java 8 is the minimum supported Java version since AEM 6.2. Also AEM 6.1 is out of support since quite a while now (https://helpx.adobe.com/support/programs/eol-matrix.html) so at the same time we should increase the minimum supported AEM version to AEM 6.2 (or newer).

tomasalmeida:
Hi, should the develop branch code be jdk 1.7 compliant? I was trying to compile the code locally and it imports some java 8 lib classes (they are not used).

kwin:
It should be according to the prerequisites in the readme. Please create a new issue to restore Java 7 compatibility.
Update: I prepared https://github.com/Netcentric/accesscontroltool/pull/414 to check API compatibility with Java 7.

PR #414, opened 2020-01-03
add animal-sniffer

kwin:
only declare versions in pluginMgmt

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -82,6 +82,7 @@
         <slf4j.version>1.7.6</slf4j.version>
         <bnd.version>4.1.0</bnd.version>
         <bouncycastle.version>1.64</bouncycastle.version>
+        <java.target.version>7</java.target.version>
     </properties>
 
     <modules>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -410,6 +411,16 @@
                     <artifactId>maven-source-plugin</artifactId>
                     <version>3.0.1</version>
                 </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-enforcer-plugin</artifactId>
+                    <version>3.0.0-M2</version>
```
tomasalmeida, line 417: A new version is available (3.0.0-M3), should this plugin be updated?

pom.xml
```diff
@@ -410,6 +411,16 @@
...
+                </plugin>
+                <plugin>
+                    <groupId>org.sonatype.plugins</groupId>
+                    <artifactId>nexus-staging-maven-plugin</artifactId>
+                    <version>1.6.3</version>
```
tomasalmeida, line 422: A new version is available (1.6.8), should this plugin be updated?

pom.xml
```diff
@@ -489,14 +505,34 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <source>1.7</source>
-                    <target>1.7</target>
+                    <source>1.${java.target.version}</source>
+                    <target>1.${java.target.version}</target>
```
tomasalmeida, line 509: in order to be able to upgrade to junit5, is it a good idea to use different jdk for unit testing and code compilation (as described here -> https://gist.github.com/aslakknutsen/9648594)?

tomasalmeida (approved):

kwin:
We will do plugin upgrades with version 3.0 (in the context of #400).

PR #417, opened 2020-01-08
Bugfix/415 416 better install hook handling

### LINKED ISSUE #415: Fix package property handling for INSTALLED vs PREPARE phase of install hook
The package property  to control if the install hook runs in INSTALLED vs PREPARE phase is wrongly implemented (inverse) (see #287). Also the property should use camel case `actool.atInstalledPhase`

### LINKED ISSUE #416: Ensure installhook only runs in non-cloud setups
Ensure install hook only runs in non-cloud setups, for cloud setups #405 is to be used.

That way the install hook can be safely included in the packages that should both work on local instances (for local testing) and in the cloud.

ghenzler:
Closes #415
Closes #416


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/installhook/AcToolInstallHook.java
```diff
@@ -52,9 +54,12 @@ public void execute(InstallContext context) throws PackageException {
     }
 
     private boolean shouldInstallInPhaseInstall(PackageProperties properties) {
-        return !Boolean.parseBoolean(properties.getProperty(PROPERTY_ACTOOL_INSTALL_AT_INSTALLED_PHASE));
```
ghenzler, line 55: it needs to be inverse here (no `!`)

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/installhook/AcToolInstallHook.java
```diff
@@ -25,7 +26,8 @@
 public class AcToolInstallHook extends OsgiAwareInstallHook {
 
     private static final Logger LOG = LoggerFactory.getLogger(AcToolInstallHook.class);
-    private static final String PROPERTY_ACTOOL_INSTALL_AT_INSTALLED_PHASE = "actool.atinstalledphase";
+    private static final String PROPERTY_ACTOOL_INSTALL_AT_INSTALLED_PHASE = "actool.atInstalledPhase";
```
ghenzler, line 29: The property didn't work as intended anyway (and we released only one version) - let's use the opportunity to use proper camel case for the property name (for readability)

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/installhook/AcToolInstallHook.java
kwin: right, good catch. We should rename the method to shouldInstallInPhaseInstalled at the same time, there is no phase INSTALL.

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/installhook/AcToolInstallHook.java
kwin: if we really want to do this, the documentation needs to be updated as well

ghenzler: yes, docu is already updated

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/installhook/AcToolInstallHook.java
ghenzler: Fixed in c6a9da28cf143ebbf43dc36e40d2056cf8efe13f

kwin (approved):

PR #420, opened 2020-01-09
#402 Fix filter rule for the case of existing directories

inedoviesov1:


accesscontroltool-apps-package/src/main/META-INF/vault/filter.xml
```diff
@@ -4,5 +4,10 @@
     <filter root="/apps/netcentric/actool/components"/>
     <filter root="/apps/netcentric/actool/content"/>
     <filter root="/apps/netcentric/actool/clientlibs"/>
-    <filter root="/apps/cq/core/content/nav/tools/security/actool"/>
+    <filter root="/apps/cq/core/content/nav">
+        <include pattern="/apps/cq/core/content/nav"/>
```
kwin, line 8: why only those 4 levels?

ghenzler: The problem is if there are `nt:unstructured` nodes, `/libs/cq/core/content/nav` is the first one to be `nt:unstructured` under libs, parents of that can be `nt:folder` or `sling:Folder` (or `sling:OrderedFolder`), that is interoperatable

kwin: but what if the parent is nt:unstructured as well? We should not make any assumptions on the repo state prior to installation?

PR #427, opened 2020-01-22
update plugins to their most recent versions

kwin:


pom.xml
```diff
@@ -403,17 +397,17 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-javadoc-plugin</artifactId>
-                    <version>3.0.1</version>
+                    <version>3.2.1</version>
```
ghenzler: latest javadoc version is 3.1.1: https://search.maven.org/artifact/org.apache.maven.plugins/maven-javadoc-plugin (running `mvn javadoc:javadoc` fails)

accesscontroltool-bundle/pom.xml
```diff
@@ -284,16 +263,18 @@
...
--exportcontents: ${packages;VERSIONED}
+# export all versioned packages except for conditional ones (https://github.com/bndtools/bnd/issues/3721#issuecomment-579026778)
+-exportcontents: ${removeall;${packages;VERSIONED};${packages;CONDITIONAL}}
 Import-Package: org.apache.jackrabbit.oak.spi.security.authentication.external;resolution:=optional;version="[1.0,3)",org.apache.jackrabbit.oak.spi.security.authentication.external.basic;resolution:=optional,com.adobe.granite.crypto;resolution:=optional,com.adobe.granite.jmx.annotation;resolution:=optional,com.day.cq.security.util;resolution:=optional,*
+# embed  el.api and jasper.el (https://github.com/bndtools/bnd/issues/3704) in expanded form
+-includeresource: @el-api-[0-9.]*.jar!/*,\
+                  @jasper-el-[0-9.]*.jar!/*
+# make build reproducible (https://github.com/bndtools/bnd/tree/master/maven/bnd-maven-plugin#reproducible-builds)
+-noextraheaders: true
+-snapshot: SNAPSHOT
```
ghenzler, line 279: If we use `SNAPSHOT` here we should also do so in https://github.com/Netcentric/accesscontroltool/blob/1c50f204c23b56705391908ec33f5f6ed3c4d9a5/accesscontroltool-startuphook-bundle/pom.xml#L107

pom.xml
kwin: Thanks, I adjusted the version in the wrong line, should be fixed now!

pom.xml
```diff
@@ -79,20 +79,14 @@
...
     </properties>
 
-    <modules>
-        <module>accesscontroltool-bundle</module>
-        <module>accesscontroltool-startuphook-bundle</module>
-        <module>accesscontroltool-apps-package</module>
-        <module>accesscontroltool-content-package</module>
-        <module>accesscontroltool-package</module>
-        <module>accesscontroltool-oakindex-package</module>
-        <module>accesscontroltool-exampleconfig-package</module>
```
ghenzler, line 94: Removing `accesscontroltool-oakindex-package` and `accesscontroltool-exampleconfig-package` will not release them anymore. Even though `oakindex-package` has rarely changed it will be confusing to have to use an arbitrary old version 2.5.1 (instead the same version as the AC tool)

accesscontroltool-apps-package/pom.xml
```diff
@@ -34,69 +34,24 @@
...
-                                    <username>${publish.crx.username}</username>
-                                    <password>${publish.crx.password}</password>
-                                </configuration>
-                            </execution>
-                        </executions>
-                    </plugin>
-                </plugins>
-            </build>
-        </profile>
-    </profiles>
```
ghenzler, line 101: Having the profiles for deployment only in the packages that are normally needed to be installed during dev cycles allowed to keep all all modules (oak index/examples in the regular module list)

ghenzler (changes requested):

pom.xml
kwin: With the release they are built: https://github.com/Netcentric/accesscontroltool/pull/427/files#diff-600376dffeb79835ede4a0b285078036R631

accesscontroltool-apps-package/pom.xml
kwin: Those are only build with dedicated profiles, so I don't think we should have a dedicated per profile per module for deployment but rather only profiles to build (and optionally deploy) the modules

accesscontroltool-bundle/pom.xml
kwin: Will move the common settings to parent, haven't seen that we have two bundles now.

kwin:
obsolete now

PR #429, opened 2020-01-23
fixes #428 removed sling:orderBefore from /apps/cq/core/content/nav/t…

### LINKED ISSUE #428: Touch UI sling:orderBefore node causes other "extra" navitems to disappear
Hi,

We just updated to version 2.5.0 of your fabulous ACTool, but we're experiencing a bug in the AEM navigation overlays.

There is a sling:orderBefore property on the /apps/cq/core/content/nav/tools/security node. This is set to "deployment".

But we also have our custom TouchUI items in there, as well as an entry of ACS AEM Commons (/apps/cq/core/content/nav/tools/acs-commons).

Issue is that both entries defined in /apps/cq/core/content/nav/tools disappear from the menu due to that property.

When I remove the sling:orderBefore property from /apps/cq/core/content/nav/tools/security it all works perfectly well. I can see our own entry plus the one from ACS AEM Commons.

I experimented with some sling:orderBefore values on our own entry, but I can't seem to find a working combination. 

I have seen this before. Likely a bug in the sling resource merger when using sling:orderBefore.

I'll create a before/after CRX package of the specific tree so you can see for yourself.

Can I suggest to remove that property to fix this issue?
And, could you give me a timeline when a new release would be cut?

Kind regards

Wim

wimsymons:
…ools/security so it won't hide other navigation bar items

### DIFF
--- a/accesscontroltool-apps-package/src/main/jcr_root/apps/cq/core/content/nav/tools/security/.content.xml
+++ b/accesscontroltool-apps-package/src/main/jcr_root/apps/cq/core/content/nav/tools/security/.content.xml
@@ -1,5 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <jcr:root xmlns:sling="http://sling.apache.org/jcr/sling/1.0" xmlns:jcr="http://www.jcp.org/jcr/1.0" xmlns:rep="internal"
-        jcr:primaryType="sling:OrderedFolder"
-        sling:orderBefore="deployment">
+        jcr:primaryType="sling:OrderedFolder">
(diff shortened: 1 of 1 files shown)


kwin (changes requested):
As outlined in https://github.com/Netcentric/accesscontroltool/issues/428#issuecomment-578715268 this will change the order with regards to a standard AEM instance. Also I fail to see the issue with that statement.

wimsymons:
This should not be fixed. It is an issue with AEM 6.4.4.0 having a buggy version of the Sling Resource Merger bundle.

PR #441, opened 2020-03-17
clarify documentation on how built-in authorizable right's are extended

kwin:


docs/Configuration.md
```diff
@@ -43,7 +43,7 @@ property | comment | required
 name | Name of the group as shown in UI. Sets the property `profile/givenName` of that group. | optional
 description | Description of the group | optional
 externalId | Required for AC setups since AEM 6.2 SP1 that synchronize groups from LDAP to AEM. The value has to be in format LDAP-DN;IDP-NAME where LDAP-DN is the full distinguished name and IDP-NAME is configured in OSGI config PID org.apache.jackrabbit.oak.security.authentication.ldap.impl.LdapIdentityProvider property "provider-name". Example: `externalId: "cn=group-name,ou=mydepart,ou=Groups,dc=comp,dc=com;IDPNAME"`. Since v1.9.3 | optional
-path | Path of the intermediate node either relative or absolute. If relative, /home/groups is automatically prefixed. By default some implementation specific path is choosen. Usually the full group path the concatenated (intermediate) path and the authorizable id  | optional
+path | Path of the intermediate node either relative or absolute. If relative, `/home/groups` is automatically prefixed. By default some implementation specific path is choosen. Usually the full group path is the (intermediate) path concatenated with the authorizable id  | optional
```
ghenzler: now that we are looking at it: `Usually the full group path is the (intermediate) path concatenated with the authorizable id` is generally wrong with recent versions, is it?  Now it's an obfuscated path instead....

kwin: Should be fixed with https://github.com/Netcentric/accesscontroltool/pull/441/commits/b8c5db1799e9213262372c93ed634f0ed45b6f3e.

Issue #467, opened 2020-06-25
NPE on biz.netcentric.cq.tools.actool.slingsettings.ExtendedSlingSettingsServiceImpl service activation


BohdanYurevych:
When service biz.netcentric.cq.tools.actool.slingsettings.ExtendedSlingSettingsServiceImpl is activated java.lang.NullPointerException: null is thrown when there are no configs for this service(no default dev/stage/prod runmodes).
AEM 6.1 SP2
ACTools 2.5.3

``` 
25.06.2020 12:04:18.597 *ERROR* [CM Event Dispatcher (Fire ConfigurationEvent: pid=biz.netcentric.cq.tools.actool.slingsettings.ExtendedSlingSettingsServiceImpl)] biz.netcentric.cq.tools.accesscontroltool.bundle [biz.netcentric.cq.tools.actool.slingsettings.ExtendedSlingSettingsServiceImpl(2874)] The activate method has thrown an exception (java.lang.NullPointerException)
java.lang.NullPointerException: null
        at java.util.Objects.requireNonNull(Objects.java:203)
        at java.util.Arrays$ArrayList.<init>(Arrays.java:3813)
        at java.util.Arrays.asList(Arrays.java:3800)
        at biz.netcentric.cq.tools.actool.slingsettings.ExtendedSlingSettingsServiceImpl.activate(ExtendedSlingSettingsServiceImpl.java:52)
        at sun.reflect.GeneratedMethodAccessor722.invoke(Unknown Source)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.apache.felix.scr.impl.helper.BaseMethod.invokeMethod(BaseMethod.java:222)
        at org.apache.felix.scr.impl.helper.BaseMethod.access$500(BaseMethod.java:37)
        at org.apache.felix.scr.impl.helper.BaseMethod$Resolved.invoke(BaseMethod.java:615)
        at org.apache.felix.scr.impl.helper.BaseMethod.invoke(BaseMethod.java:499)
```

PR #468, opened 2020-06-25
467: added check for null runmodes in case no config is provided and …

BohdanYurevych:
…no default runmode (dev/stage/prod) #467

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/slingsettings/ExtendedSlingSettingsServiceImpl.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/slingsettings/ExtendedSlingSettingsServiceImpl.java
@@ -49,8 +49,12 @@ public class ExtendedSlingSettingsServiceImpl implements ExtendedSlingSettingsSe
         if(isCloudReady) {
             extendedRunmodes.add(ADDITIONAL_RUNMODE_CLOUD);
         }
-        List<String> additionalRunmodes = Arrays.asList(config.additionalRunmodes());
-        extendedRunmodes.addAll(additionalRunmodes);
+
+        if (config.additionalRunmodes() != null) {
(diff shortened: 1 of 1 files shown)


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/slingsettings/ExtendedSlingSettingsServiceImpl.java
```diff
@@ -49,8 +49,12 @@ public void activate(Config config) {
         if(isCloudReady) {
             extendedRunmodes.add(ADDITIONAL_RUNMODE_CLOUD);
         }
-        List<String> additionalRunmodes = Arrays.asList(config.additionalRunmodes());
-        extendedRunmodes.addAll(additionalRunmodes);
+
+        if (config.additionalRunmodes() != null) {
+            List<String> additionalRunmodes = Arrays.asList(config.additionalRunmodes());
+            extendedRunmodes.addAll(additionalRunmodes);
+        }
```
ghenzler, line 56: The stacktrace shows that `config.additionalRunmodes()` returns null even though in theory, the annotation interface https://github.com/Netcentric/accesscontroltool/blob/789f7d5aab9cdfe0864b6033a68e74173745131a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/slingsettings/ExtendedSlingSettingsServiceImpl.java#L34 defines a default value and therefore the value should never be null.

But anyway, the stacktrace shows this can be  a problem and this PR should fix it, merging it.

kwin: It can be null in DS 1.3 in case no entry has been provided. This has been fixed in DS 1.4 (https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html#d0e46608) which will now instead return an empty array

PR #472, opened 2020-06-29
Don't use sub service name for service user mapping

### LINKED ISSUE #471: Wrong Sling ServiceUserMapper configuration for JcrInstallUrlHandler
In https://github.com/Netcentric/accesscontroltool/blob/develop/accesscontroltool-package/src/main/jcr_root/apps/netcentric/actool/config/org.apache.sling.serviceusermapping.impl.ServiceUserMapperImpl.amended-ncActoolSystemUser.config there is a user mapping defined for `BSN`:`biz.netcentric.cq.tools.accesscontroltool.bundle` and `subServiceName`:`actool`.

This does not apply everywhere as in some places in the source code where service resource resolver or system session is retrieved `null` is used as `subServiceName`.

This is https://github.com/Netcentric/accesscontroltool/search?q=loginService&unscoped_q=loginService and https://github.com/Netcentric/accesscontroltool/search?q=getServiceResourceResolver&unscoped_q=getServiceResourceResolver.

Therefore I am always getting failures like `org.apache.sling.api.resource.LoginException: Cannot derive user name for bundle biz.netcentric.cq.tools.accesscontroltool.bundle [591] and sub service null at org.apache.sling.resourceresolver.impl.ResourceResolverFactoryImpl.getServiceResourceResolver(ResourceResolverFactoryImpl.java:79) at biz.netcentric.cq.tools.actool.impl.JcrInstallUrlHandler.activate(JcrInstallUrlHandler.java:39) at ...`

kwin:
This fixes #471

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/aceinstaller/AceBeanInstallerIncremental.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/aceinstaller/AceBeanInstallerIncremental.java
@@ -258,7 +258,7 @@ public class AceBeanInstallerIncremental extends BaseAceBeanInstaller implements
             Session newSession = null;
             try {
 
-                newSession = slingRepository.loginService(Constants.USER_AC_SERVICE, null);
+                newSession = slingRepository.loginService(null, null);
                 Session relevantSessionToUse;
                 if (newSession.nodeExists(origAceBean.getJcrPath())) {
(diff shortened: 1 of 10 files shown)


ghenzler (changes requested):
Let's just use the subServiceName in JcrInstallUrlHandler as well

kwin:
Let‘s only introduce subservice names once we need them. Not before! Are you worried about introducing regressions with this?

ghenzler (approved):

PR #490, opened 2020-10-15
#489 Allow variables with arbitrary yaml list/map structures

ghenzler:
Closes #489

PR #491, opened 2020-10-16
#257 allow to reuse variables from other yaml files

### LINKED ISSUE #257: Allow to reuse variables from other YAML files
Currently the functional logic (loops, variables, ...) only work on a per YAML-file basis. It is not possible to declare variables once in one YAML file and use it then in multiple others. This is often desirable though for more complex setups.

Currently the code in https://github.com/Netcentric/accesscontroltool/blob/65f4dfd962a525a0ac1ea2e5f4433f07c5a5eed6/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/configreader/YamlMacroProcessorImpl.java#L59 leads to the fact that for each YAML file the variables map is being empty.

In addition one would need to clarify the order in which YAML files are processed (depth first vs. breadth first or all in alphabetical order).

ghenzler:
Closes #257

### DIFF
--- a/accesscontroltool-bundle/src/test/java/biz/netcentric/cq/tools/actool/configreader/YamlMacroProcessorTest.java
+++ b/accesscontroltool-bundle/src/test/java/biz/netcentric/cq/tools/actool/configreader/YamlMacroProcessorTest.java
@@ -12,6 +12,7 @@ import static biz.netcentric.cq.tools.actool.configreader.YamlConfigReaderTest.g
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.MockitoAnnotations.initMocks;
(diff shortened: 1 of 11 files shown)


ghenzler:
includes #490 and should be merged after

accesscontroltool-bundle/src/test/java/biz/netcentric/cq/tools/actool/configreader/YamlMacroProcessorTest.java
```diff
@@ -481,5 +477,75 @@ public void testDefWithUnderscore() throws Exception {
...
+        for(int i=0; i<expectedVals.length && it.hasNext(); i++) {
+            AceBean aceBean = it.next();
+            assertEquals("Expected path for bean "+i, expectedVals[i][0], aceBean.getJcrPath());
+            assertEquals("Expected initial content for bean "+i, expectedVals[i][1], aceBean.getInitialContent());
+        }
+    
+
+    }
+    
 }
```
kwin, line 581: Can we have a test for global variables as well?

ghenzler: Added one in a1feaba

docs/AdvancedFeatures.md
```diff
@@ -215,19 +218,21 @@ There is also a multi-line variant of the DEF statement that allows to define co
       - name: "Value ${simpleMap[loopVar]}"
 ```
 
+### Global variables 
+
 By default, variables are local, this means the scope of a variable is limited to:
  * the lines in the very same yaml file following the definition till it is either redefined or the end of the yaml file is reached 
  * `FOR` loop in which variable is defined or re-defined.
```
kwin, line 225: Please add a sentence about precedence of variables (I guess same name local variable takes precedence) and also add tests for same name local and global variables

ghenzler: See 1fec561

kwin (approved):

Issue #492, opened 2020-10-16
Add Sonarcloud.io integration


kwin:
The project at https://sonarcloud.io/dashboard?id=Netcentric_accesscontroltool should be populated by the GitHub CI/CD actioin.

PR #502, opened 2020-10-19
#492 Sonar fixes

ghenzler:


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/AcToolUiService.java
```diff
@@ -73,7 +73,7 @@ protected void doPost(final HttpServletRequest req, final HttpServletResponse re
 
         RequestParameters reqParams = RequestParameters.fromRequest(req, acInstallationService);
         LOG.info("Received POST request to apply AC Tool config with configurationRootPath={} basePaths={}",
-                reqParams.configurationRootPath, reqParams.basePaths);
+                escapeHtml4(reqParams.configurationRootPath), reqParams.basePaths);
```
kwin: why do you escape for HTML parsers in a log? A log line is never evaluated as HTML IMHO

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/AcToolUiService.java
```diff
@@ -85,7 +85,7 @@ protected void doPost(final HttpServletRequest req, final HttpServletResponse re
         resp.setContentType("text/plain");
         if (((PersistableInstallationLogger) log).isSuccess()) {
             resp.setStatus(HttpServletResponse.SC_OK);
-            pw.println("Applied AC Tool config from " + reqParams.configurationRootPath + ":\n" + msg);
+            pw.println("Applied AC Tool config from " + escapeHtml4(reqParams.configurationRootPath) + ":\n" + msg);
```
kwin: This is plain text and therefore should not contain any HTML escaping!

kwin (changes requested):
I think the Sonarcloud issues are rather false positives and should just be disabled with annotations (https://github.com/SonarSource/sonar-java/blob/2e3b7d25cebac51d5040d30dc3048c2f575d6e92/its/plugin/projects/suppress-warnings/src/main/java/SuppressWarningsAtMethodLevelWithRule.java#L4)

kwin:
Still one FP: https://sonarcloud.io/project/issues?id=Netcentric_accesscontroltool&open=AXVCFo2RKPqjJP9f4pZJ&pullRequest=502&resolved=false&types=VULNERABILITY

Let us ignore javasecurity:S5145 as well.

ghenzler:
@kwin For some reason Sonar still reports on `javasecurity:S5145` even thought it should be suppressed via 7e8ed5456c8cdf8b6debf5584eacf7dbc9cce5dc

PR #504, opened 2020-10-19
#503 - proposal to speed up group analysis

### LINKED ISSUE #503: Improve the performance of applying groups and users (group_config / user_config)
If a setup has a large number of groups and users (in particular the number of users can be large like > 100,000), the execution time of the AC Tool can get fairly long. If the AC tool is triggered by the install hook that is included in a package, the installation using the package manager might run into a timeout:

```
[INFO] Package uploaded, now installing...
[INFO] ERROR: HTTP call to http://localhost:4503/crx/packmgr/service/console.html/etc/packages/amg/package-1.0.0-SNAPSHOT.zip?cmd=install&recursive=true failed: Read timed out (consider to increase the socket timeout using -Dvault.httpSocketTimeoutSec)
[INFO] ---------------
[INFO] HTTP call failed, try again (1/24) after 5 second(s)...
```

Increasing the timeout can solve the problem, but in some special cases that is not enough. 

One of the bottlenecks is `AuthorizableInstallerServiceImpl.getDeclaredMembersWithoutRegularUsers` which analyses each member of the group linearly. A solution with jdk8+ would be using a parallel stream to speed up this process, using jdk7, I believe we need to handle the threads manually.

tomasalmeida:
analyse each member of a group in a separate thread for performance

once actool requires jdk8, this class can be dropped in a parallel stream can replace it.

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/NotRegularMembers.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/NotRegularMembers.java
@@ -1,0 +1,93 @@
+package biz.netcentric.cq.tools.actool.authorizableinstaller.impl;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+import javax.jcr.RepositoryException;
+
+import org.apache.jackrabbit.api.security.user.Authorizable;
+
+import biz.netcentric.cq.tools.actool.helper.Constants;
+
+/**
+ * Auxiliar class to store a list of non regular members.
+ */
+class NotRegularMembers {
+
+    private final List<String> members = new CopyOnWriteArrayList<>();
+    private final ExecutorService executor = Executors.newFixedThreadPool(10);
+
+    private RepositoryException thrownException;
+    private final Object thrownExceptionLocker = new Object();
+
+    /**
+     * Add the given member to the set if it is considered a non regular user, discard otherwise
+     *
+     * @param member candidate to be added
+     */
+    public void addOrDiscard(final Authorizable member) {
+        if (executor.isShutdown()) {
+            throw new IllegalStateException("Set is closed and new member cannot be added");
+        }
+        final MemberCheckerWorker memberCheckerWorker = new MemberCheckerWorker(member);
+        executor.execute(memberCheckerWorker);
+    }
+
+    /**
+     * Get the resultant set of filtered users
+     *
+     * @return the set of id of the users
+     * @throws RepositoryException in case it could not verify the id of some user
+     */
+    public Set<String> getSet() throws RepositoryException {
+        waitAllMemberCheckersToFinish();
+        throwExceptionIfAnyWorkerFailed();
+        return new HashSet<>(members);
+    }
+
+    private void throwExceptionIfAnyWorkerFailed() throws RepositoryException {
+        if (thrownException != null) {
+            throw thrownException;
+        }
+    }
+
+    private void waitAllMemberCheckersToFinish() {
+        executor.shutdown();
+        while (!executor.isTerminated()) {
+        }
+    }
+
+    private class MemberCheckerWorker implements Runnable {
+        private final Authorizable member;
+
+        MemberCheckerWorker(final Authorizable member) {
+            this.member = Objects.requireNonNull(member);
+        }
+
+        @Override
+        public void run() {
+            try {
+                if (member != null && !isRegularUser(member)) {
+                    members.add(member.getID());
+                }
+            } catch (final RepositoryException e) {
+                synchronized (thrownExceptionLocker) {
+                    if (thrownException == null) {
+                        thrownException = e;
+                    }
+                }
+            }
+        }
+
+        private boolean isRegularUser(final Authorizable member) throws RepositoryException {
+            return !member.isGroup() // if user
+                    && !member.getPath().startsWith(Constants.USERS_ROOT + "/system/") // but not system user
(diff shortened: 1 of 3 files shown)


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/NotRegularMembers.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.jackrabbit.api.security.user.Authorizable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import biz.netcentric.cq.tools.actool.helper.Constants;
+
+/**
+ * Auxiliar class to store a list of non regular members.
+ */
+class NotRegularMembers {
```
kwin, line 23: what are non-regular members? Please clarify in javadoc.

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/NotRegularMembers.java
```diff
@@ -0,0 +1,103 @@
...
+
+        MemberCheckerWorker(final Authorizable member) {
+            this.member = Objects.requireNonNull(member);
+        }
+
+        @Override
+        public void run() {
+            try {
+                if (member != null && !isRegularUser(member)) {
+                    members.add(member.getID());
```
kwin, line 86: is this really the bottleneck?

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/NotRegularMembers.java
ghenzler: maybe a better name would be to use a positive name, e.g. `GroupsAndSystemUserMembers`

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/NotRegularMembers.java
ghenzler: I think @tomasalmeida  did a performance test and this supposedly helps, but I also don't understand why (I would have thought `installedGroup.getDeclaredMembers()` is the bottleneck). Even if the query in getDeclaredMembers() was lazy, `currentMemberInRepo.hasNext()` is also called in one thread only. @tomasalmeida do you have more insight what exactly is slow? (is it `member.getPath()` in `isRegularUser` below?)

kwin: let us enable this only via a feature toggle and not by default. Then we can test some more configurations with this option before rolling it out to everyone.

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/NotRegularMembers.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+
+    private void waitAllMemberCheckersToFinish() {
+        executor.shutdown();
+        while (!executor.isTerminated()) {
+            try {
+                executor.awaitTermination(10, TimeUnit.SECONDS);
+            } catch (final InterruptedException e) {
+                LOG.error("Thread executor await termination was interrupted.", e);
```
kwin, line 70: You should never swallow this exception: https://www.ibm.com/developerworks/library/j-jtp05236/

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/NotRegularMembers.java
tomasalmeida: Sure thing! I will add a sling feature flag for that, and perform more tests :-)

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/NotRegularMembers.java
kwin: You also have to consider https://issues.apache.org/jira/browse/OAK-2609 though.

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/NotRegularMembers.java
kwin: A simple flag in the YAML config is imho best in this case.

kwin:
For some reason Sonar seems to understand only the syntax 

> Use @SuppressWarnings(‘cs.RULENAME, cs.OTHERULE’) to specify multiple rules, separating each new rule with a comma.

(https://docs.codescan.io/hc/en-us/articles/360012109711-Ignoring-violations)

tomasalmeida:
I will close this PR and rethink the solution as you suggested, I open a new one with all the points hightlighted here.

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/NotRegularMembers.java
tomasalmeida: we tested more scenarios yesterday and `installedGroup.getDeclaredMembers()` is indeed a bottleneck. It takes 8ms to get each member multiplied by our 100k users (which derives to 13 minutes). We are testing if instead of using this method, we can loop using the group property, but we are not certain of the result.

PR #505, opened 2020-10-19
#455 Allow to configure impersonators of users

### LINKED ISSUE #455: Provide impersonation configuration
I have a lot of groups, and for testing I need testusers for each group for easier validation.

But on the other hand side I need everyone to login with a personalized account, so I cannot provide these users with the capability to login, but the users should use the impersonalization feature to impersonate into one of these testusers.

(To ease handling, I create these testusers with a secret password providing just the hash value in the yaml files).

I haven't found a way to create these testusers with the ability that all users can impersonate into them.

ghenzler:
Closes #455


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/ImpersonationInstallerServiceImpl.java
```diff
@@ -0,0 +1,118 @@
...
+        }
+    }
+
+    private void applyChanges(User user, InstallationLogger installationLog, List<String> impersonatorsToBeSet, Impersonation impersonation,
+            List<String> currentImpersonators) throws RepositoryException {
+        List<String> impersonatorsToAdd = new ArrayList<>();
+        impersonatorsToAdd.addAll(impersonatorsToBeSet);
+        impersonatorsToAdd.removeAll(currentImpersonators);
+        Iterator<String> impersonatorsToAddIt = impersonatorsToAdd.iterator();
+        while(impersonatorsToAddIt.hasNext()) {
```
kwin: some whitespace missing, are you using the right formatter?

docs/Configuration.md
```diff
@@ -98,10 +98,10 @@ disabled | Can be set to `true` or an arbitrary reason string to disable a user.
 profileContent | Allows to provide [enhanced docview xml](https://jackrabbit.apache.org/filevault/docview.html) that will reset the profile to the given structure after each run | optional
 preferencesContent | Allows to provide [enhanced docview xml](https://jackrabbit.apache.org/filevault/docview.html) that will reset the preferences node to the given structure after each run | optional
 socialContent | Allows to provide [enhanced docview xml](https://jackrabbit.apache.org/filevault/docview.html) that will reset the social node to the given structure after each run | optional
+impersonationAllowedFor | List of users that may impersonate as this user. May be provided as yaml list or comma-separated list (String). If this property is omitted the impersonators of the user remain untouched, otherwise impersonators are added as given in config (and existing ones not configured are removed). To force an empty list of impersonators, an empty yaml list may be used: `impersonationAllowedFor: []`. | optional
```
kwin: why comma separated list? Just requires additional effort with parsing and would require escaping. Let us stick to plain yaml at least for new features.

kwin (changes requested):

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/ImpersonationInstallerServiceImpl.java
ghenzler: didn't have formatting on save turned on, fixed in 2e37da80c88ad3cf677c8025834df868d282e1a5

docs/Configuration.md
ghenzler: I'd rather stick with the pattern we have - we can remove comma-support and other yaml quirks we we have in yaml format with version 3.0.0 IMHO

kwin: Then at least make it clearer that comma-separated lists are the deprecated variant!

ghenzler: that is fine, ec644ff98fd6c85

kwin (approved):

PR #510, opened 2020-10-22
Fix section identifier in documentation

martinloeffler:
Current example throws an invalid section identifier exception.

### DIFF
--- a/docs/AdvancedFeatures.md
+++ b/docs/AdvancedFeatures.md
@@ -256,7 +256,7 @@ It is possible to define global variables (that are available across multiple ya
 
 ```
 - global_config:
-     vars:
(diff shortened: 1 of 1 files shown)


kwin:
For me it works without `vars` being an array. As there are never multiple `vars` fields, an array would not make sense either. @martinloeffler You are probably suffering from another issue here.

martinloeffler:
> For me it works without `vars` being an array. As there are never multiple `vars` fields, an array would not make sense either. @martinloeffler You are probably suffering from another issue here.

Sorry, my bad 🤦 I must have messed up the indentation when copying over the example.
Thanks for pointing it out @kwin. I will delete the PR.

PR #516, opened 2020-11-13
Support placeholders in YAML scalars which are resolved by

### LINKED ISSUE #515: Support interpolation of YAML values with Felix Configadmin Plugin Interpolation
Currently [CryptoSupport](https://experienceleague.adobe.com/docs/experience-manager-64/administering/security/encryption-support-for-configuration-properties.html?lang=en#decryption-support) is used for symmetrically encrypting both passwords with the instance's master key. Other than that there is no means to protect sensitive values like passwords or private keys.

This does no longer work well with AEMaaCS, as all environments have different master keys. The new mechanism to access sensitive information is via [Secret Configuration Values](https://experienceleague.adobe.com/docs/experience-manager-cloud-service/implementing/deploying/configuring-osgi.html?lang=en#secret-configuration-values). That mechanism is currently only used via [Felix Configadmin Plugin Interpolation](https://github.com/apache/felix-dev/blob/master/configadmin-plugins/interpolation/README.md) but the same approach could also be used outside of OSGi configuration admin.

Unfortunately Adobe does not expose the path of those secret files nor their encoding, but both can be evaluated by leveraging the Framework Properties `org.apache.felix.configadmin.plugin.interpolation.secretsdir` and `org.apache.felix.configadmin.plugin.interpolation.file.encoding` (<https://github.com/AdobeDocs/cloudmanager-api-docs/issues/26#issuecomment-726169270>). Obviously this requires that the referenced secrets in the yaml are set via Cloud Manager API before.

kwin:
org.apache.felix.configadmin.plugin.interpolation

This closes #515

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/configreader/YamlConfigurationMerger.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/configreader/YamlConfigurationMerger.java
@@ -19,13 +19,16 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Pattern;
 
 import javax.jcr.RepositoryException;
 import javax.jcr.Session;
 
 import org.apache.commons.lang3.time.StopWatch;
+import org.osgi.service.cm.ConfigurationPlugin;
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Reference;
+import org.osgi.service.component.annotations.ReferenceCardinality;
 import org.osgi.service.component.annotations.ReferencePolicyOption;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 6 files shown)


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/configreader/YamlConfigurationMerger.java
```diff
@@ -92,8 +101,15 @@ public AcConfiguration getMergedConfigurations(
...
         final Set<String> obsoleteAuthorizables = new HashSet<String>();
 
-        final Yaml yamlParser = new Yaml();
-
+        final Yaml yamlParser;
+        if (interpolationPlugin != null) {
+            yamlParser = new Yaml(new YamlConfigurationAdminPluginScalarConstructor(installLog, interpolationPlugin));
+            // bind constructor to certain scalar formats (compare with https://bitbucket.org/asomov/snakeyaml/src/master/src/test/java/org/yaml/snakeyaml/env/EnvVariableTest.java)
+            yamlParser.addImplicitResolver(YamlConfigurationAdminPluginScalarConstructor.TAG, CONFIG_ADMIN_INTERPOLATOR_FORMAT, "$[" );
+            installLog.addMessage(LOG, "Using YAML parser with ConfigurationAdmin Plugin placeholder support");
```
ghenzler, line 109: nice to see this approach, wasn't aware the parser can be extended that way

ghenzler (approved):

PR #518, opened 2020-11-15
Allow non-encrypted PKCS#8 private keys

### LINKED ISSUE #517: Support non-encrypted private keys
Currently private keys are only supported in encrypted PKCS#8 format (https://github.com/Netcentric/accesscontroltool/blob/develop/docs/Configuration.md#configuration-of-keys). For some use cases unencrypted ones work just as well (e.g. for local dev environments with read-only access) or when using interpolated security values (#515).

kwin:
Also allow privatePassword to be clear text.
Both is useful together with value interpolation.

This closes #517 #519


docs/Configuration.md
```diff
@@ -171,8 +171,8 @@ Each key entry in the `keys` section stands for a key alias in the key store. Th
...
+privatePassword | The password for decrypting the encrypted private key. The password itself can be encrypted with the AEM Crypto Support (i.e. encrypted with the AEM master key of the according instance). Alternatively one can rely on value interpolation for this value. Once the key is added to the keystore this password is no longer relevant as there the private key is encrypted with the password of the AEM keystore itself. | no
```
ghenzler: > ...no longer relevant as **there**

should be  `then`  instead of `there`?

pom.xml
```diff
@@ -269,12 +269,7 @@
                 <groupId>org.bouncycastle</groupId>
                 <artifactId>bcpkix-jdk15on</artifactId>
                 <version>${bouncycastle.version}</version>
-            </dependency>
-            <!-- must be direct dependency to be embedded -->
-            <dependency>
-                <artifactId>bcprov-jdk15on</artifactId>
-                <groupId>org.bouncycastle</groupId>
-                <version>${bouncycastle.version}</version>
+                <scope>provided</scope>
```
ghenzler, line 272: So `bcpkix-jdk15on` we would always embed, and `bcprov-jdk15on` users only install if they need it?

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/aem/AemCryptoSupport.java
```diff
@@ -9,8 +9,15 @@
...
 
 /** OSGi service to encapulate AEM crypto support (to make it optional) */
-public interface AemCryptoSupport {
+public abstract class AemCryptoSupport {
 
-    public String unprotect(String password);
+    public abstract String unprotect(String password);
+
+    /**
+     * Copied from CryptoImpl, need it separately to be able to check if decryption is necessary before CryptoSupport is loaded
```
ghenzler: Looking at this I first thought it wouldn't work because AemCryptoSupport does not become available if the dependency CryptoSupport is not there. Now I see it is a static method and hence it should work. I think it's confusing though to make the interface an abstract class just to allow for this static method. I would rather create a class `AemCryptoSupportHelper` with a sole static method `isProtected()`, then it's at least clear what is happening and `AemCryptoSupport` can remain an interface

pom.xml
kwin: No, both need to be installed (we don't embed anything). Please compare with the documentation.

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/aem/AemCryptoSupport.java
kwin: I don't get the concern here: Coming up with a separate class just to place one static method in it feels too heavy. What is you concern of converting the interface to an abstract class?
(Since Java 8 even interfaces are allowed to have static methods, but we are yet bound to Java 7)

ghenzler: I think it's better for maintainability to have this separated (an extra class does not cost anything and clearly separates the "decoupling of AEM crypto support" aspect from the utility method that is always needed (leaving AemCryptoSupport as interface makes that easier)

pom.xml
ghenzler: With current version 2.6.2 we embed both bcpkix and bcprov, don't we break backwards compatibility if this needs to be installed in a different way now?

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/aem/AemCryptoSupport.java
kwin: For me those two belong together. The utilty method needs to mimic exactly what the CryptoSupportImpl is doing (otherwise it is less useful)

> an extra class does not cost anything

It makes it harder to see where it belongs to

ghenzler (approved):

PR #533, opened 2021-01-04
#532 Only change password if it has actually changed

### LINKED ISSUE #532: Do not change password if password has not actually changed
For each user (usually test users) from the AC Tool config that contains a password, a line as the following is printed in the log 

```
04.01.2021 16:44:31.076 *INFO* [qtp328160521-192-ACTool-Config-Worker] com.adobe.granite.security.user.internal.audit.AuditAuthorizableAction Password for User 'tu-global-editor' was changed
```

For setups with many tests users, there are many changes. Also it's not really nice to change the password if it is actually just changed to the same value. 

The reason why the password is updated on every run today, is that the password is obviously not stored clear text and hence it's not a simple string compare with the repo value to check if the password has changed (see https://github.com/Netcentric/accesscontroltool/blob/523a38435a84cc83266ada96d3641b6b549ac56b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthorizableInstallerServiceImpl.java#L233 for the code)

However it is easy enough to rely on repo.login() to check if the currently set password works.

ghenzler:
Closes #532

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthorizableInstallerServiceImpl.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthorizableInstallerServiceImpl.java
@@ -25,6 +25,7 @@ import java.util.regex.Pattern;
 import javax.jcr.Node;
 import javax.jcr.RepositoryException;
 import javax.jcr.Session;
+import javax.jcr.SimpleCredentials;
 import javax.jcr.UnsupportedRepositoryOperationException;
 import javax.jcr.ValueFactory;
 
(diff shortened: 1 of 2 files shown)


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthorizableInstallerServiceImpl.java
```diff
@@ -226,11 +231,25 @@ private void installKeys(Map<String, Key> keys, String userId, ResourceResolver
...
     void setUserPassword(final AuthorizableConfigBean authorizableConfigBean,
-                             final User authorizableToInstall) throws RepositoryException, AuthorizableCreatorException {
+            final User authorizableToInstall, InstallationLogger installLog) throws RepositoryException, AuthorizableCreatorException {
+
+        String userId = authorizableToInstall.getID();
         String password = getPassword(authorizableConfigBean);
-        authorizableToInstall.changePassword(password);
+        Session sessionForUser = null;
+        try {
+            sessionForUser = repository.login(new SimpleCredentials(userId, password.toCharArray()));
```
ghenzler, line 241: If it was possible to use the oak security provider (as in https://github.com/apache/jackrabbit-oak/blob/15e2befce204f0a3b7457777d912db06c93d235d/oak-core/src/main/java/org/apache/jackrabbit/oak/core/ContentRepositoryImpl.java#L161-L163) directly it would be more light-weight, to me it seems not easily possible though...

kwin: Oak Core is internal API and subject to change, so the approach with JCR API is a good approach.

kwin (approved):

PR #534, opened 2021-01-05
#503 Improve the performance of applying groups and users

### LINKED ISSUE #503: Improve the performance of applying groups and users (group_config / user_config)

ghenzler:
Closes #503

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/PrefetchedAuthorizablesUserManager.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/PrefetchedAuthorizablesUserManager.java
@@ -1,0 +1,124 @@
+package biz.netcentric.cq.tools.actool.authorizableinstaller.impl;
+
+import static biz.netcentric.cq.tools.actool.history.PersistableInstallationLogger.msHumanReadable;
+
+import java.security.Principal;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+
+import javax.jcr.RepositoryException;
+
+import org.apache.jackrabbit.api.security.user.Authorizable;
+import org.apache.jackrabbit.api.security.user.Group;
+import org.apache.jackrabbit.api.security.user.Query;
+import org.apache.jackrabbit.api.security.user.QueryBuilder;
+import org.apache.jackrabbit.api.security.user.User;
+import org.apache.jackrabbit.api.security.user.UserManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import biz.netcentric.cq.tools.actool.history.InstallationLogger;
+
+/** Prefetches all authorizables in constructor to be able to answer getAuthorizable(userId) requests much quicker than the ootb oak
+ * UserManager itself. */
+class PrefetchedAuthorizablesUserManager implements UserManager {
+
+    private static final Logger LOG = LoggerFactory.getLogger(PrefetchedAuthorizablesUserManager.class);
+
+    private final UserManager delegate;
+    private final Map<String, Authorizable> authorizableCache;
+
+    public PrefetchedAuthorizablesUserManager(UserManager delegate, InstallationLogger installLog) throws RepositoryException {
+        this.delegate = delegate;
+        this.authorizableCache = new HashMap<>();
+
+        long startPrefetch = System.currentTimeMillis();
+        Iterator<Authorizable> allAuthorizables = delegate.findAuthorizables(new Query() {
+            public <T> void build(QueryBuilder<T> builder) {
+                // fetch all authorizables
+            }
+        });
+
+        while (allAuthorizables.hasNext()) {
+            Authorizable auth = allAuthorizables.next();
+            authorizableCache.put(auth.getID(), auth);
+        }
+
+        installLog.addMessage(LOG, "Prefetched " + authorizableCache.size() + " authorizables in "
+                + msHumanReadable(System.currentTimeMillis() - startPrefetch));
+    }
+
+    @Override
+    public Authorizable getAuthorizable(String id) throws RepositoryException {
+        Authorizable authorizable = authorizableCache.get(id);
+        if (authorizable == null) {
+            authorizable = delegate.getAuthorizable(id);
+            authorizableCache.put(id, authorizable);
+        }
+        return authorizable;
+    }
+
+    public int getCacheSize() {
+        return authorizableCache.size();
+    }
+
+    // --- delegated methods
+
+    public Authorizable getAuthorizable(Principal principal) throws RepositoryException {
+        return delegate.getAuthorizable(principal);
+    }
+
+    public Authorizable getAuthorizableByPath(String path) throws RepositoryException {
+        return delegate.getAuthorizableByPath(path);
+    }
+
+    public Iterator<Authorizable> findAuthorizables(String relPath, String value) throws RepositoryException {
+        return delegate.findAuthorizables(relPath, value);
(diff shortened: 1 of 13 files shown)


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/PrefetchedAuthorizablesUserManager.java
```diff
@@ -0,0 +1,125 @@
...
+
+    private final UserManager delegate;
+    private final Map<String, Authorizable> authorizableCache;
+
+    public PrefetchedAuthorizablesUserManager(UserManager delegate, InstallationLogger installLog) throws RepositoryException {
+        this.delegate = delegate;
+        this.authorizableCache = new HashMap<>();
+
+        long startPrefetch = System.currentTimeMillis();
+        Iterator<Authorizable> allAuthorizables = delegate.findAuthorizables(ALL_AUTHORIZABLES_QUERY);
```
ghenzler: Prefetching all authorizables (using one query under the hood) turns out to be a lot more efficient than running many individual getAuthorizable() calls

kwin: Do we know why? I once figured out in https://projects.netcentric.biz/wiki/display/CQ/Authorizable+vs.+Principal
> UserManager.getAuthorizable(String id)
This method is usually very fast because no query is involved because the node's jcr:uuid property (i.e. the nodes ID) is equal to UUID.nameUUIDFromBytes(name.getBytes(StandardCharsets.UTF_8)).
Therefore, Session.getNodeByIdentifier(...)is used to quickly retrieve the underlying node.
That is the case for all authorizables being created via the UserManager!

Is that observation no longer correct?

ghenzler: So I just did measurements with a ~4000 authorizables setup and caching it like this brings down the installation time significantly. However I found that my current project's performance was improved but  @tomasalmeida 's problem  was not yet solved (~100.000 users in getDeclaredMembers()) - hence I followed this path further, see https://github.com/Netcentric/accesscontroltool/blob/1413e9315c2d3199e8f52dda087def95c39c3802/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/PrefetchingUserManager.java#L40

tomasalmeida: I deployed this code and tested it against an old version. Comparison values:

`time <deploy command>`

actool 2.3.2 
real	2m43.451s
real	2m45.496s
real	3m5.596s
real	2m57.520s

this branch (installed over the 2.3.2)
real	0m22.456s
real	0m18.718s
real	0m18.224s

I have in my local server more than 150000 users.

thank you so much for the implementation and the improvement 👏

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/PrefetchingUserManager.java
```diff
@@ -0,0 +1,196 @@
...
+
+    private final Map<String, Set<String>> nonRegularUserMembersByAuthorizableId = new HashMap<>();
+    private final Map<String, Set<String>> isMemberOfByAuthorizableId = new HashMap<>();
+
+    public PrefetchingUserManager(UserManager delegate, final ValueFactory valueFactory, InstallationLogger installLog)
+            throws RepositoryException {
+        this.delegate = delegate;
+
+        long startPrefetch = System.currentTimeMillis();
+        Iterator<Authorizable> allGroupsAndSystemUsersAndAnonymous = delegate.findAuthorizables(new Query() {
```
kwin: `authorizablesToPrefetch` instead of `allGroupsAndSystemUsersAndAnonymous`?

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/PrefetchingUserManager.java
```diff
@@ -0,0 +1,196 @@
...
+ * <p>
+ * The membership relationships are cached in lookup maps to quickly be able to query the repository state for both
+ * authorizable.declaredMemberOf() (this we call in this class also) and group.getDeclaredMembers(). Having called declaredMemberOf() for
+ * all groups of the repository has also retrieved all memberships (expect regular user memberships), hence the method getDeclaredMembers()
+ * does not have to be called anymore. This is particularly useful because that way the AC tool does not have to iterate over a potentially
+ * huge number of users in production to then only filter out a few relevant groups in the end (the AC Tool does not touch user memberships
+ * to groups).
+ * </p>
+ */
+class PrefetchingUserManager implements UserManager {
```
kwin: UserManager is a provider-type (https://github.com/apache/jackrabbit-oak/blob/trunk/oak-jackrabbit-api/src/main/java/org/apache/jackrabbit/api/security/user/UserManager.java), therefore this implementation is prone to break with future upgrades....

Since Jackrabbit/Oak doesn't provide a delegate class I would rather put this logic in a dedicated class which does not implement UserManager and only implement those methods which are actually called.

kwin (changes requested):

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/PrefetchingUserManager.java
ghenzler: Fair enough, see https://github.com/Netcentric/accesscontroltool/blob/3b967710b9b09979b0c4bfdee41123f7175da0d0/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthInstallerUserManager.java#L13 to ensure new methods in UserManager will not break the AC tool in future

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/PrefetchingUserManager.java
ghenzler: done

kwin (approved):

tomasalmeida (approved):

PR #548, opened 2021-01-26
Startup hook should implement SlingRepositoryInitializer to propagate

### LINKED ISSUE #545: AEMaaCS: Startup Hook executed too early for ACLs on mutable content
Currently the Install Hook is disabled by default in AEMaaCS (https://github.com/Netcentric/accesscontroltool/blob/48e1bfee12347ac958288bffce0ff0c978d77ff9/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/installhook/AcToolInstallHook.java#L80).

When the Startup Hook is executed during the Docker build (when first starting the instance) it works on top of the already set up immutable repo (being initialized with https://github.com/apache/sling-org-apache-sling-jcr-packageinit/blob/master/src/main/java/org/apache/sling/jcr/packageinit/impl/ExecutionPlanRepoInitializer.java). That works fine.

When the Startup Hook is executed while the new Kubernetes pod is starting, the mutable content packages are not yet installed (i.e. the content to which to apply the ACLs might not be there yet), so this execution might fail during the first deployment (but works then on subsequent ones).

kwin:
exceptions properly

This closes #545 partially


accesscontroltool-startuphook-bundle/src/main/java/biz/netcentric/cq/tools/actool/startuphook/impl/AcToolStartupHookServiceImpl.java
```diff
@@ -177,4 +146,28 @@ private void copyAcHistoryToOrFromApps(boolean isCloudReady) {
...
+                activationMode);
+
+        boolean applyOnStartup = (activationMode == Config.StartupHookActivation.ALWAYS) 
+                || (isCloudReady && activationMode == Config.StartupHookActivation.CLOUD_ONLY);
+
+        if(applyOnStartup) {
+            List<String> relevantPathsForInstallation = getRelevantPathsForInstallation(repo);
+            LOG.info("Running AcTool with "
+                    + (relevantPathsForInstallation.isEmpty() ? "all paths" : "paths " + relevantPathsForInstallation) + "...");
+            acInstallationService.apply(null, relevantPathsForInstallation.toArray(new String[relevantPathsForInstallation.size()]),
```
ghenzler, line 163: @kwin during the repo initialisation process this method participates, IIRC the `SlingRepository` will not be available as service yet? (see https://github.com/Netcentric/accesscontroltool/blob/b8f5453ce9fed0ac953466d05d08da0c790a7248/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/impl/AcInstallationServiceImpl.java#L103)

kwin: right, this probably requires some refactoring of `AcInstallationServiceImpl()`

Issue #554, opened 2021-02-01
unmanagedAcePathsRegex feature slows down ACL Tool execution


insideaem:
Hi there,
I am currently working on a project where we use the ACT Tool with lots of yaml files. We've noticed that since the introduction of the unmanagedAcePathsRegex feature, the tool now takes about 5.8h to run instead of about 1h.

Number of YAML files: 1322
Number of Authorizables: 10206
Number of ACEs: 47329
Number of paths: 15001
Runtime: 5.8h

30.10.2020 10:30:46.581 *INFO* biz.netcentric.cq.tools.actool.impl.AcInstallationServiceImpl *** Starting installation of 10206 authorizables from configuration...
30.10.2020 11:55:40.070 *INFO* biz.netcentric.cq.tools.actool.impl.AcInstallationServiceImpl Finished installation of authorizables without errors in 1.4h
30.10.2020 12:16:42.962 *INFO* biz.netcentric.cq.tools.actool.impl.AcInstallationServiceImpl *** Starting installation of 47329 ACE configurations for 15001 paths in content nodes using strategy AceBeanInstallerIncremental...
30.10.2020 16:16:33.165 *INFO* biz.netcentric.cq.tools.actool.aceinstaller.BaseAceBeanInstaller ACL Update Statistics: Changed=0 Unchanged=11023 Path not found=3978 (action cache hit/miss=28144/12)
30.10.2020 16:16:33.166 *INFO* biz.netcentric.cq.tools.actool.aceinstaller.BaseAceBeanInstaller *** Finished installation of 15001 ACLs in 4h
30.10.2020 16:16:33.201 *INFO* biz.netcentric.cq.tools.actool.impl.AcInstallationServiceImpl Successfully applied AC Tool configuration in 5.8h
--

We've analyzed it and traced the error back to the call of **getAuthorizableConfigByPrincipalName** in the method **removeUnmanagedPrincipalNamesAtPath**.

Our current solution is to fork the AC Tool and add an OSGI configuration to the AclInstallationServiceImpl to be able to limit the check of the unmanagedAcePathsRegex only to certain authorizable we know we are using the feature in our yaml files.
In our case we only use it for everyone.

I will submit a PR with our changes that you can integrate. Or in case you have a better idea please feel free to implement.

ghenzler:
[...]

PR #555, opened 2021-02-01
[554] Configuration for the authorizables allowed to use the unmanage…

insideaem:
…d path feature in order to improve performance
This is to fix issue #554


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/impl/AcInstallationServiceImpl.java
```diff
@@ -144,6 +146,9 @@
         
         @AttributeDefinition(name="Configuration path(s)", description="JCR path(s) where the config files reside (usually it's just one, can be multiple for multitenant setups)")
         String[] configurationRootPaths() default {};
+        
+        @AttributeDefinition(name="Allowed principals for unmanaged path(s)", description="List of principals for which the unmanaged paths should be checked")
+        String[] allowedPrincipalsForUnmanagedPaths() default {};
```
ghenzler, line 151: Introducing a new config property is something we would want to avoid (also if we introduced one, it should be part of the global config section, the strategy is to have `configurationRootPaths` as the only OSGi config prop, `intermediateSaves` has been removed in the latest version since 3.x

ghenzler:
Using the simpler approach https://github.com/Netcentric/accesscontroltool/pull/590 for now

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/impl/AcInstallationServiceImpl.java
insideaem: ok

PR #558, opened 2021-02-02
Improve resilience when search index does not (yet) contain the groups

### LINKED ISSUE #557: NPE in AuthInstallerUserManagerPrefetchingImpl in AEMCloud with incomplete/invalid search index in place
With the most recent 2.7.2-SNAPSHOT I ran into the following exception

```
*INFO* [qtp9431062-822-ACTool-Config-Worker] biz.netcentric.cq.tools.actool.authorizableinstaller.impl.AuthInstallerUserManagerPrefetchingImpl Prefetched 1 authorizables in 15ms
...
*ERROR* [qtp9431062-822] biz.netcentric.cq.tools.actool.installhook.AcToolInstallHook Exception during execution of install hook: biz.netcentric.cq.tools.actool.authorizableinstaller.AuthorizableCreatorException: java.lang.NullPointerException
...
Caused by: biz.netcentric.cq.tools.actool.authorizableinstaller.AuthorizableCreatorException: java.lang.NullPointerException
	at biz.netcentric.cq.tools.actool.impl.AcInstallationServiceImpl.installAuthorizables(AcInstallationServiceImpl.java:564)
	at biz.netcentric.cq.tools.actool.impl.AcInstallationServiceImpl.installAcConfiguration(AcInstallationServiceImpl.java:343)
	at biz.netcentric.cq.tools.actool.impl.AcInstallationServiceImpl.installMergedConfigurations(AcInstallationServiceImpl.java:668)
	at biz.netcentric.cq.tools.actool.impl.AcInstallationServiceImpl.installConfigurationFiles(AcInstallationServiceImpl.java:304)
	at biz.netcentric.cq.tools.actool.impl.AcInstallationServiceImpl.installConfigurationFiles(AcInstallationServiceImpl.java:270)
	at biz.netcentric.cq.tools.actool.installhook.impl.AcToolInstallHookServiceImpl.installYamlFilesFromPackage(AcToolInstallHookServiceImpl.java:63)
	at biz.netcentric.cq.tools.actool.installhook.AcToolInstallHook.install(AcToolInstallHook.java:99)
	... 73 common frames omitted
Caused by: java.lang.NullPointerException: null
	at biz.netcentric.cq.tools.actool.authorizableinstaller.impl.AuthInstallerUserManagerPrefetchingImpl.<init>(AuthInstallerUserManagerPrefetchingImpl.java:94)
	at biz.netcentric.cq.tools.actool.authorizableinstaller.impl.AuthorizableInstallerServiceImpl.installAuthorizables(AuthorizableInstallerServiceImpl.java:114)
	at biz.netcentric.cq.tools.actool.impl.AcInstallationServiceImpl.installAuthorizables(AcInstallationServiceImpl.java:551)
	... 79 common frames omitted

```

kwin:
This closes #557

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthInstallerUserManagerPrefetchingImpl.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthInstallerUserManagerPrefetchingImpl.java
@@ -61,6 +61,7 @@ class AuthInstallerUserManagerPrefetchingImpl implements AuthInstallerUserManage
 
         long startPrefetch = System.currentTimeMillis();
         Iterator<Authorizable> authorizablesToPrefetchIt = delegate.findAuthorizables(new Query() {
+            // all groups and user "anonymous"
             public <T> void build(QueryBuilder<T> builder) {
                 builder.setCondition(
                         builder.or( //
@@ -91,7 +92,15 @@ class AuthInstallerUserManagerPrefetchingImpl implements AuthInstallerUserManage
                 Group memberOfGroup = declaredMemberOf.next();
                 String memberOfGroupId = memberOfGroup.getID();
                 isMemberOfByAuthorizableId.get(authId).add(memberOfGroupId);
-                nonRegularUserMembersByAuthorizableId.get(memberOfGroupId).add(authId);
+                final Set<String> nonRegularUserMembers;
+                if (!nonRegularUserMembersByAuthorizableId.containsKey(memberOfGroupId)) {
+                    // in case the group was not found by the previous query
+                    nonRegularUserMembers = new HashSet<String>();
+                    nonRegularUserMembersByAuthorizableId.put(memberOfGroupId, nonRegularUserMembers);
+                } else {
+                    nonRegularUserMembers = nonRegularUserMembersByAuthorizableId.get(memberOfGroupId);
+                }
+                nonRegularUserMembers.add(authId);
                 membershipCount++;
             }
         }
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthInstallerUserManagerPrefetchingImpl.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthInstallerUserManagerPrefetchingImpl.java
@@ -61,12 +61,13 @@ class AuthInstallerUserManagerPrefetchingImpl implements AuthInstallerUserManage
 
         long startPrefetch = System.currentTimeMillis();
         Iterator<Authorizable> authorizablesToPrefetchIt = delegate.findAuthorizables(new Query() {
-            // all groups and user "anonymous"
             public <T> void build(QueryBuilder<T> builder) {
                 builder.setCondition(
-                        builder.or( //
+                        builder.or( 
+                                // this is true for groups and system users
                                 builder.neq("@" + JcrConstants.JCR_PRIMARYTYPE, valueFactory.createValue(UserConstants.NT_REP_USER)),
-                                builder.eq("@" + UserConstants.REP_AUTHORIZABLE_ID, valueFactory.createValue(Constants.USER_ANONYMOUS))) //
+                                // this is true for user anonymous
+                                builder.eq("@" + UserConstants.REP_AUTHORIZABLE_ID, valueFactory.createValue(Constants.USER_ANONYMOUS)))
                 );
             }
         });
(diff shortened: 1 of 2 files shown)


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthInstallerUserManagerPrefetchingImpl.java
```diff
@@ -61,6 +61,7 @@ public AuthInstallerUserManagerPrefetchingImpl(UserManager delegate, final Value
 
         long startPrefetch = System.currentTimeMillis();
         Iterator<Authorizable> authorizablesToPrefetchIt = delegate.findAuthorizables(new Query() {
+            // all groups and user "anonymous"
```
ghenzler: this comment is fully correct, system users are also included

kwin: Fixed in https://github.com/Netcentric/accesscontroltool/pull/558/commits/52b654000e5b9db417285444e52b9b8a23851ea7.

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthInstallerUserManagerPrefetchingImpl.java
```diff
@@ -91,7 +93,15 @@ public AuthInstallerUserManagerPrefetchingImpl(UserManager delegate, final Value
...
-                nonRegularUserMembersByAuthorizableId.get(memberOfGroupId).add(authId);
+                final Set<String> nonRegularUserMembers;
+                if (!nonRegularUserMembersByAuthorizableId.containsKey(memberOfGroupId)) {
+                    // in case the group was not found by the previous query
+                    nonRegularUserMembers = new HashSet<String>();
+                    nonRegularUserMembersByAuthorizableId.put(memberOfGroupId, nonRegularUserMembers);
+                } else {
+                    nonRegularUserMembers = nonRegularUserMembersByAuthorizableId.get(memberOfGroupId);
+                }
+                nonRegularUserMembers.add(authId);
```
tomasalmeida, line 104: as this code is self contained, may be it will be easier to read if we separate it in a new method

```
    private void addMemberToAuthId(final String memberOfGroupId, final String authId) {
        final Set<String> nonRegularUserMembers;
        if (!nonRegularUserMembersByAuthorizableId.containsKey(memberOfGroupId)) {
            // in case the group was not found by the previous query
            nonRegularUserMembers = new HashSet<String>();
            nonRegularUserMembersByAuthorizableId.put(memberOfGroupId, nonRegularUserMembers);
        } else {
            nonRegularUserMembers = nonRegularUserMembersByAuthorizableId.get(memberOfGroupId);
        }
        nonRegularUserMembers.add(authId);
    }
```

and then we can call it here like 

```addMemberToAuthId(memberOfGroupId, authId);```

**Review by tomasalmeida — APPROVED** [review]

File: accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthInstallerUserManagerPrefetchingImpl.java [hunk 3]
```diff
@@ -91,7 +93,15 @@ public AuthInstallerUserManagerPrefetchingImpl(UserManager delegate, final Value
                 Group memberOfGroup = declaredMemberOf.next();
                 String memberOfGroupId = memberOfGroup.getID();
                 isMemberOfByAuthorizableId.get(authId).add(memberOfGroupId);
-                nonRegularUserMembersByAuthorizableId.get(memberOfGroupId).add(authId);
+                final Set<String> nonRegularUserMembers;
+                if (!nonRegularUserMembersByAuthorizableId.containsKey(memberOfGroupId)) {
```
ghenzler, line 97: this condition is very problematic, because the authorizable installer relies on having a full/complete result of the methods `getDeclaredIsMemberOf()` und `getDeclaredMembersWithoutRegularUsers()` - if the query from the `QueryBuilder` above is not 100% reliable, we should switch to traversal of /home/groups and /home/users/system (having these two roots does not require to traverse over all users fortunately). The traversal could be fallback or default (if performance is similar to the query)

kwin: Not the if condition is problematic but the unreliable search. I am also fine with throwing another exception to make it more obvious that the index is corrupt/incomplete in this case, but this cannot always be detected reliably. I hope we can agree on the fact that running into an NPE is a situation we should prevent....

ghenzler: let's use https://github.com/Netcentric/accesscontroltool/pull/617/files instead then

Issue #562, opened 2021-03-06
Renaming group with member results in "underlying tree has been disconnected"


jgorgolewski:
Since version 2.7.1 installations that try to rename group using migrateFrom feature fail with following error:
```
Migrating from group old-group-name to new-group-name
- Deleted group old-group-name
E Could not process yaml files (biz.netcentric.cq.tools.actool.authorizableinstaller.AuthorizableCreatorException: java.lang.IllegalStateException: Authorizable old-group-name: underlying tree has been disconnected.)
Exception while installing configurations: biz.netcentric.cq.tools.actool.authorizableinstaller.AuthorizableCreatorException: java.lang.IllegalStateException: Authorizable old-group-name: underlying tree has been disconnected. 
```
The issue happens when renamed group has member managed by actool. For example, when moving from this configuration:
```
-
  group_config:
    -
      old-group-name:
        -
          name: old-group-name
    -
      group-name-approvers:
        -
          name: group-name-approvers
          isMemberOf: old-group-name
```
to this configuration:
```
-
  group_config:
    -
      new-group-name:
        -
          name: new-group-name
          migrateFrom: old-group-name
    -
      group-name-approvers:
        -
          name: group-name-approvers
          isMemberOf: new-group-name
```

Issue seems to be similar to #542 (side effect of #534) but happens also on 2.7.2. Likely related to  AuthInstallerUserManagerPrefetchingImpl is not clearing caches when authorizables are removed.

Following details are present in the logs:
```
biz.netcentric.cq.tools.actool.authorizableinstaller.impl.AuthorizableInstallerServiceImpl - start installation of authorizable: new-group-name
biz.netcentric.cq.tools.actool.authorizableinstaller.impl.AuthorizableInstallerServiceImpl start adding new-group-name to assignedGroups
biz.netcentric.cq.tools.actool.authorizableinstaller.impl.AuthorizableInstallerServiceImpl Successfully created new group: new-group-name
biz.netcentric.cq.tools.actool.authorizableinstaller.impl.AuthorizableInstallerServiceImpl Migrating from group old-group-name to new-group-name
biz.netcentric.cq.tools.actool.authorizableinstaller.impl.AuthorizableInstallerServiceImpl - Deleted group old-group-name
biz.netcentric.cq.tools.actool.authorizableinstaller.impl.AuthorizableInstallerServiceImpl - start installation of authorizable: group-name-approvers
[...]

PR #564, opened 2021-03-12
#562 - if the authorizable is removed, caches are cleared

tomasalmeida:
When we remove an authorizable, if the cache is not refreshed, it will return outdated relations which cause the error seen.

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthInstallerUserManagerPrefetchingImpl.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthInstallerUserManagerPrefetchingImpl.java
@@ -8,6 +8,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 import javax.jcr.RepositoryException;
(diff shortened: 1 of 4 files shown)


kwin:
> Should I go on?

Probably not a good idea then, also because implementing an interface of Jackrabbit API which is meant as provider type leads to a very tight coupling with a high chance of breaking with future versions. Let us go with your original approach.

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthInstallerUserManagerPrefetchingImpl.java
```diff
@@ -140,6 +140,35 @@ private void refreshAuthorizableCacheIfNeeded(final String id, final Authorizabl
...
+
+        // remove auth from cache
+        authorizableCache.remove(authorizable.getID());
+        // if auth is a group, remove auth from the cache which lists groups of a given user
+        removeAuthorizableFromGroupsCache(authorizable);
+
+        authorizable.remove();
+    }
+
+    private void removeAuthorizableFromGroupsCache(final Authorizable authorizable) throws RepositoryException {
```
kwin: I would rather name `removeGroup`, use `Group` as an argument type and do the instanceof check in the `removeAuthorizable`

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthInstallerUserManagerPrefetchingImpl.java
```diff
@@ -140,6 +140,35 @@ private void refreshAuthorizableCacheIfNeeded(final String id, final Authorizabl
         }
     }
 
+    @Override
+    public void removeAuthorizable(final Authorizable authorizable) throws RepositoryException {
+        if (authorizable == null) {
```
kwin: Is this ever called with `null`? I would rather go for throwing an exception in that case to not hide a programming error.

kwin (commented):
LGTM in general, only minor nitpicks.

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthInstallerUserManagerPrefetchingImpl.java
tomasalmeida: done!

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthInstallerUserManagerPrefetchingImpl.java
tomasalmeida: done!

kwin (approved):

Issue #573, opened 2021-04-23
NPE in AEMaaCS when IF macro condition evaluates to NULL


ofarganc:
When testing a condition that evaluates to null, the ACTool throws the following exception.

Example of condition: if ${site["jcr:content"]["isCommerce"]}
This could be because of missing content or other reasons.

`22.04.2021 17:01:39.489 [cm-p33328-e114754-aem-author-66bff65877-j4c5x] *ERROR* [FelixStartLevel] biz.netcentric.cq.tools.actool.impl.AcInstallationServiceImpl Exception in AceServiceImpl: {}
java.lang.NullPointerException: null
	at biz.netcentric.cq.tools.actool.configreader.YamlMacroProcessorImpl.evaluateIfStatement(YamlMacroProcessorImpl.java:241) [biz.netcentric.cq.tools.accesscontroltool.bundle:2.8.0]
	at biz.netcentric.cq.tools.actool.configreader.YamlMacroProcessorImpl.transform(YamlMacroProcessorImpl.java:142) [biz.netcentric.cq.tools.accesscontroltool.bundle:2.8.0]`

PR #574, opened 2021-04-23
bugix-573 Prevent NPE when IF condition evaluates to NULL

ofarganc:
Fixes
https://github.com/Netcentric/accesscontroltool/issues/573

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/configreader/YamlMacroProcessorImpl.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/configreader/YamlMacroProcessorImpl.java
@@ -8,17 +8,8 @@
  */
 package biz.netcentric.cq.tools.actool.configreader;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.LinkedHashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.Map.Entry;
-import java.util.Set;
 import java.util.regex.Matcher;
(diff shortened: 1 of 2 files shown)


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/configreader/YamlMacroProcessorImpl.java
```diff
@@ -8,17 +8,8 @@
...
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.LinkedHashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
ghenzler: Let's leave the imports untouched (to keep the formatting/code style consistent with the rest of the project)

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/configreader/YamlMacroProcessorImpl.java
```diff
@@ -238,7 +229,12 @@ private Object evaluateIfStatement(Map<String, Object> variables, Object objVal,
             InstallationLogger installLog, Session session) {
         String condition = ifMatcher.group(1).trim();
 
-        boolean expressionIsTrue = elEvaluator.evaluateEl(condition, Boolean.class, variables);
+        Boolean expressionIsTrue = elEvaluator.evaluateEl(condition, Boolean.class, variables);
+
+        if (expressionIsTrue == null) {
+            LOG.warn("Expression {} evaluates to null, returning false", condition);
```
ghenzler: we should use the methods in installLog for logging (they log both to the error.log and to the AC Tool log available in the system console)

ghenzler (approved):

Issue #575, opened 2021-04-29
Exception when adding externalId and changing id for an existing group


phornig:
With 2.8.0 I get an exception when trying to add an externalId when also changing the id (changing some letters to upper case) of an existing group :

```
29.04.2021 14:53:01.948 *INFO* [OsgiInstallerImpl] biz.netcentric.cq.tools.actool.installhook.impl.OsgiAwareInstallHook Exception while installing configurations: biz.netcentric.cq.tools.actool.authorizableinstaller.AuthorizableCreatorException: java.lang.Ille
galStateException: Authorizable content-editor: underlying tree has been disconnected.
29.04.2021 14:53:01.948 *WARN* [OsgiInstallerImpl] org.apache.jackrabbit.vault.packaging.impl.InstallHookProcessorImpl Hook actool threw package exception. PREPARE aborted.
org.apache.jackrabbit.vault.packaging.PackageException: java.lang.IllegalStateException: Authorizable content-editor: underlying tree has been disconnected.
        at biz.netcentric.cq.tools.actool.installhook.AcToolInstallHook.install(AcToolInstallHook.java:106) [biz.netcentric.cq.tools.accesscontroltool.bundle:2.8.0]
        at biz.netcentric.cq.tools.actool.installhook.AcToolInstallHook.execute(AcToolInstallHook.java:42) [biz.netcentric.cq.tools.accesscontroltool.bundle:2.8.0]
        at org.apache.jackrabbit.vault.packaging.impl.InstallHookProcessorImpl.execute(InstallHookProcessorImpl.java:143) [org.apache.jackrabbit.vault:3.4.0]
        at org.apache.jackrabbit.vault.packaging.impl.ZipVaultPackage.extract(ZipVaultPackage.java:226) [org.apache.jackrabbit.vault:3.4.0]
        at org.apache.jackrabbit.vault.packaging.impl.JcrPackageImpl.extract(JcrPackageImpl.java:401) [org.apache.jackrabbit.vault:3.4.0]
        at org.apache.jackrabbit.vault.packaging.impl.JcrPackageImpl.extract(JcrPackageImpl.java:360) [org.apache.jackrabbit.vault:3.4.0]
        at org.apache.jackrabbit.vault.packaging.impl.JcrPackageImpl.install(JcrPackageImpl.java:354) [org.apache.jackrabbit.vault:3.4.0]
[...]

PR #576, opened 2021-05-04
Always use lower case for the the auth cache key to match Jackrabbits…

phornig:
… default UserManagerImpl case-insensitive handling of authorizable id's, #575

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthInstallerUserManagerPrefetchingImpl.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthInstallerUserManagerPrefetchingImpl.java
@@ -39,6 +39,10 @@ import biz.netcentric.cq.tools.actool.history.InstallationLogger;
  * upon creation.
  * </p>
  * <p>
+ * Authorizables as well as membership relationships are cached by their case-insensitive IDs in alignment
+ * with {@link org.apache.jackrabbit.oak.security.user.AuthorizableBaseProvider#getContentID(String authorizableId)}.
+ * </p>
+ * <p>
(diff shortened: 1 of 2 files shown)


kwin:
We already depend on Commons Collections, so why not using https://commons.apache.org/proper/commons-collections/javadocs/api-3.2.2/org/apache/commons/collections/map/CaseInsensitiveMap.html?

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthInstallerUserManagerPrefetchingImpl.java
```diff
@@ -51,7 +52,7 @@
     private static final Logger LOG = LoggerFactory.getLogger(AuthInstallerUserManagerPrefetchingImpl.class);
 
     private final UserManager delegate;
-    private final Map<String, Authorizable> authorizableCache = new HashMap<>();
+    private final Map<String, Authorizable> authorizableCache = new CaseInsensitiveMap();
```
kwin, line 59: what about the other maps below?

Also a reference towards https://github.com/apache/jackrabbit-oak/blob/cde907088a5c460178fd714cb4edc141cd45e23d/oak-core/src/main/java/org/apache/jackrabbit/oak/security/user/AuthorizableBaseProvider.java#L84 in some comments helps to understand the issue better

phornig: I don't know if the other maps need an adjustment. They don't seem to have the same issue since they don't fall back to the delegate behavior. Would you recommend to change those to CaseInsensitiveMap as well? Any other adjustments needed?

kwin (approved):
LGTM, thanks.

ghenzler (approved):

PR #580, opened 2021-05-31
Drop compatibility with AEM < 6.4

### LINKED ISSUE #400: Upgrade to Java 8 (drop support of AEM 6.1)

### LINKED ISSUE #487: Require AEM 6.4
Only AEM 6.4 and above is not yet EOL. We should drop support of older versions to be able to clean up the dependencies.

kwin:
Require Java 8
Update testing dependencies

This closes #487
This closes #400


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/api/InstallationLog.java
```diff
@@ -8,35 +8,26 @@
...
-import biz.netcentric.cq.tools.actool.installationhistory.AcInstallationHistoryPojo;
+import org.osgi.annotation.versioning.ProviderType;
 
 /**
  * Access to log messages being emitted
  *
  */
-@ConsumerType
-public interface InstallationLog extends AcInstallationHistoryPojo {
+@ProviderType
```
kwin, line 19: not sure about this change. Is it reasonable for consumers to implement this?

ghenzler: no, the change makes sense to me (was wrong in the first place)

ghenzler (approved):

PR #600, opened 2021-10-07
clarifies deprecation

nc-richard-meier:


nc-richard-meier:
Clarifies deprecation comment in spec of isMemberOf property.

ghenzler (approved):

Issue #596, opened 2021-08-31
Vulnerability with tomcat-jasper-el


marcjubero:
I’ve added `accesscontroltool-bundle (v3.0.2)` as a dependency in the project I’m working and since then, jenkins build fails due to some vulnerability related with `tomcat-jasper-el-8.5.66.jar `

Here's the jenkins error:
```
[ERROR] Failed to execute goal org.owasp:dependency-check-maven:5.2.4:check (default) on project contenthub-aem-core:
[ERROR]
[ERROR] One or more dependencies were identified with vulnerabilities:
[ERROR]
[ERROR] tomcat-jasper-el-8.5.66.jar: CVE-2021-33037, CVE-2020-8022
[ERROR]
[ERROR] See the dependency-check report for more details.
[ERROR]
```

kwin:
The latter is a false positive (CVE-2020-8022) which I reported now upstream at https://github.com/jeremylong/DependencyCheck/issues/3661. The former (https://nvd.nist.gov/vuln/detail/CVE-2021-33037) really affects 8.5.66 as well, so we should upgrade to 8.5.70.

PR #603, opened 2021-10-15
Reduce dependencies

### LINKED ISSUE #602: Reduce dependencies for AEMaaCS
For planned changes in AEMaaCS build process (in particular the image build) the required dependencies should be reduced, in particular the following changes should be made: 
* JSP API can be embedded
* Sling Commons Scheduler can be made optional (https://github.com/Netcentric/accesscontroltool/blob/develop/docs/ApplyConfig.md#upload-listener-service is rarely used and discouraged anyway, for up-and-running AEMaaCS instances it will still work but not for /apps and /libs)

ghenzler:
Closes #602

### DIFF
--- a/accesscontroltool-bundle/bnd.bnd
+++ b/accesscontroltool-bundle/bnd.bnd
@@ -3,7 +3,7 @@ Bundle-SymbolicName: biz.netcentric.cq.tools.accesscontroltool.bundle
 # export all versioned packages by default
 -exportcontents: ${packages;VERSIONED}
 # Embed (https://groups.google.com/g/bndtools-users/c/NAdU0_X8OKY)
--includeresource: tomcat-el-api-[0-9.]*.jar;lib:=true,tomcat-jasper-el-[0-9.]*.jar;lib:=true,snakeyaml-[0-9.]*.jar;lib:=true
+-includeresource: tomcat-el-api-[0-9.]*.jar;lib:=true,tomcat-jasper-el-[0-9.]*.jar;lib:=true,snakeyaml-[0-9.]*.jar;lib:=true,jsp-api-[0-9.]*.jar;lib:=true
 # 
 # allow to run in Sling without AEM bundles
 # allow to run without bouncycastle which is only necessary for some edge cases when managing keys
@@ -13,4 +13,6 @@ com.adobe.granite.keystore;resolution:=optional,\
 com.adobe.granite.jmx.annotation;resolution:=optional,\
 com.day.cq.security.util;resolution:=optional,\
 org.bouncycastle.*;resolution:=optional,\
+org.apache.sling.commons.scheduler.*;resolution:=optional,\
+org.apache.jackrabbit.oak.spi.security.principal;version="[1.5.0,3)",\
 *
--- a/accesscontroltool-bundle/bnd.bnd
+++ b/accesscontroltool-bundle/bnd.bnd
@@ -3,10 +3,7 @@ Bundle-SymbolicName: biz.netcentric.cq.tools.accesscontroltool.bundle
 # export all versioned packages by default
 -exportcontents: ${packages;VERSIONED}
 # Embed (https://groups.google.com/g/bndtools-users/c/NAdU0_X8OKY)
--includeresource: tomcat-el-api-[0-9.]*.jar;lib:=true,tomcat-jasper-el-[0-9.]*.jar;lib:=true,snakeyaml-[0-9.]*.jar;lib:=true,tomcat-jsp-api-[0-9.]*.jar;lib:=true
-
-# avoid the dependency to servlet API coming in via tomcat-jsp-api
--contract: !JavaServlet,*
+-includeresource: tomcat-el-api-[0-9.]*.jar;lib:=true,tomcat-jasper-el-[0-9.]*.jar;lib:=true,snakeyaml-[0-9.]*.jar;lib:=true
 
 # allow to run in Sling without AEM bundles
 # allow to run without bouncycastle which is only necessary for some edge cases when managing keys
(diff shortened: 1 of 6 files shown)


accesscontroltool-bundle/bnd.bnd
```diff
@@ -13,4 +13,6 @@ com.adobe.granite.keystore;resolution:=optional,\
 com.adobe.granite.jmx.annotation;resolution:=optional,\
 com.day.cq.security.util;resolution:=optional,\
 org.bouncycastle.*;resolution:=optional,\
+org.apache.sling.commons.scheduler.*;resolution:=optional,\
```
kwin, line 16: we might need to reconsider optional imports: https://github.com/adobe/aem-core-wcm-components/issues/1872

kwin (changes requested):
We should adjust https://github.com/Netcentric/accesscontroltool/blob/develop/target-osgi-environment/maximum-environment/pom.xml to check that it runs without those dependencies!

accesscontroltool-bundle/pom.xml
```diff
@@ -106,7 +106,15 @@
             <artifactId>bcpkix-jdk15on</artifactId>
             <optional>true</optional>
         </dependency>
-        
+
+        <!-- JSP API to be embedded directly -->
+        <dependency>
+          <groupId>org.apache.tomcat</groupId>
+          <artifactId>jsp-api</artifactId>
+          <version>6.0.14</version>
```
kwin: dependency-checks considers this version vulnerable

accesscontroltool-bundle/bnd.bnd
ghenzler: I see, also interesting the discussion on [Felix mailing list](https://lists.apache.org/thread.html/r20cceffd651904eda7e6512b2699b5381ab8044e352bb27cb14e9964%40%3Cusers.felix.apache.org%3E) - for `org.apache.sling.commons.scheduler` it should work reliably though, in AEM it starts at start level 19 (before the AC Tool is wired), in Sling it comes slightly later ([start level 20](https://github.com/apache/sling-org-apache-sling-starter/blob/5e9f0fc352d0a9ea855409d4a90283f4397f006b/src/main/features/base.json#L70), yet it should still be ok)

accesscontroltool-bundle/pom.xml
ghenzler: this is currently using the same version as https://github.com/apache/sling-org-apache-sling-scripting-jsp-api/blob/8c36e309e7eae7277882445727733b400026de59/pom.xml#L48 - checking if I can get a better version without vulnerability

kwin: by embedding it we have to take care of that vulnerabilities don't affect us, if just using the provided one Adobe does that for us

ghenzler: 0a18191 fixes this and uses the latest dependency (the group id changed from `jsp-api` to `tomcat-jsp-api`, that wasn't immediately obvious)

ghenzler:
> We should adjust https://github.com/Netcentric/accesscontroltool/blob/develop/target-osgi-environment/maximum-environment/pom.xml to check that it runs without those dependencies!

d30e5bf removes scripting dependencies in min environment (the max env didn't include them)

accesscontroltool-bundle/pom.xml
ghenzler: turns out that the current version still has one vulnerability left: CVE-2020-8022

kwin: This is a false-positive: https://github.com/Netcentric/accesscontroltool/issues/596#issuecomment-922831860

accesscontroltool-bundle/bnd.bnd
```diff
@@ -3,8 +3,11 @@ Bundle-SymbolicName: biz.netcentric.cq.tools.accesscontroltool.bundle
 # export all versioned packages by default
 -exportcontents: ${packages;VERSIONED}
 # Embed (https://groups.google.com/g/bndtools-users/c/NAdU0_X8OKY)
--includeresource: tomcat-el-api-[0-9.]*.jar;lib:=true,tomcat-jasper-el-[0-9.]*.jar;lib:=true,snakeyaml-[0-9.]*.jar;lib:=true
-# 
+-includeresource: tomcat-el-api-[0-9.]*.jar;lib:=true,tomcat-jasper-el-[0-9.]*.jar;lib:=true,snakeyaml-[0-9.]*.jar;lib:=true,tomcat-jsp-api-[0-9.]*.jar;lib:=true
+
+# avoid the dependency to servlet API coming in via tomcat-jsp-api
+-contract: !JavaServlet,*
```
kwin: AEM 6.5.x only provides support for Servlet Spec 3.1 while Tomcat 9 is for Servlet Spec 4.0 (https://tomcat.apache.org/whichversion.html), we should therefore stick to Tomcat 8, the require capability is there for a reason in Tomcat 9!

accesscontroltool-bundle/pom.xml
```diff
@@ -22,9 +22,13 @@
     <packaging>jar</packaging>
     <name>Access Control Tool Bundle</name>
     <description>The OSGi bundle containing all classes of the Access Control Tool.</description>
+    
+    <properties>
+        <tomcat.libs.version>10.0.12</tomcat.libs.version>
```
kwin: Currently documentation states that we support EL 2.2 (https://github.com/Netcentric/accesscontroltool/blob/develop/docs/AdvancedFeatures.md#expressions). What has changed in 4.0? Is that really fully backwards compatible?

accesscontroltool-bundle/bnd.bnd
ghenzler: we currently don't have the `Require-Capability` for the servlet API, the header comes from the embedded dependency (that is only embedded because fo the EL functionality)

kwin: we depend on this dependency having the contract, so transitively we also depend on the capability!

accesscontroltool-bundle/pom.xml
ghenzler: The unit tests pass... if even, only edge cases could cause a problem. 

Also more generally, as EL is used broadly in a huge amount of systems, I would assume new versions are carefully tested to be backwards-compatible.

kwin: We need at least an update of the documentation that we now support EL 4.0 with the according link.

ghenzler: Updated doc in ba2634b

accesscontroltool-bundle/pom.xml
kwin: No longer relevant, as JSP dependency has  been removed.

kwin (approved):

PR #611, opened 2021-11-08
Support explicitly configured key store password

### LINKED ISSUE #560: Extend ACTool to set an explicit keystore password
Basically API handles the **keystorePassword** transparently (i.e. consumer does not need to pass it to any of the API calls) but there are some some implementations which still require it to be explicitly configured (again, although it is already contained in the JCR):

- For example **com.adobe.granite.jetty.ssl.internal.GraniteSslConnectorFactory**  which requires the configuration of the keystore password in OSGi config (duplicates the keystorePassword JCR property).

In this case i would like to have possibility to set explicitly keystorepassword to fully automate configuration for such implementations.

kwin:
This closes #560

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -1,21 +1,5 @@
 # Configuration File Format
 
-* [Overview](#overview)
-* [Storage of configurations in JCR](#storage-of-configurations-in-jcr)
-* [Run modes](#run-modes)
-* [Overall structure a of an AC configuration file](#overall-structure-a-of-an-ac-configuration-file)
-* [Configuration of groups](#configuration-of-groups)
-  * [Group migration](#group-migration)
-* [Configuration of users](#configuration-of-users)
-  * [Configuration of keys](#configuration-of-keys)
-    * [Creation of key pair](#creation-of-key-pair)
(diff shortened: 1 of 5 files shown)


docs/Configuration.md
```diff
@@ -1,21 +1,5 @@
...
-  * [Group migration](#group-migration)
-* [Configuration of users](#configuration-of-users)
-  * [Configuration of keys](#configuration-of-keys)
-    * [Creation of key pair](#creation-of-key-pair)
-    * [Encrypt private key](#encrypt-private-key)
-* [Configuration of ACEs](#configuration-of-aces)
-* [Global Configuration](#global-configuration)
-* [Validation](#validation)
-
-<!--- This table of contents has been generated with https://github.com/ekalinin/github-markdown-toc#gh-md-toc -->
```
ghenzler, line 17: Did you really want to remove the TOC here?

kwin: Yes, GitHub renders that automatically now: https://github.blog/changelog/2021-04-13-table-of-contents-support-in-markdown-files/

ghenzler (approved):

PR #613, opened 2021-11-12
#612 save execution marker hash immediately

### LINKED ISSUE #612: Avoid duplicate execution in startup hook
The code in https://github.com/Netcentric/accesscontroltool/blob/2e2727f304256fead1049be72665c0b0287304dd/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/impl/AcConfigChangeTracker.java#L25 can potentially cause a race condition if multiple pods are started at the same time in AEMaaCS:

image

ghenzler:
Closes #612

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/impl/AcConfigChangeTracker.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/impl/AcConfigChangeTracker.java
@@ -41,7 +41,7 @@ public class AcConfigChangeTracker {
                 }
                 
                 acHistoryNode.setProperty(executionKey, hashOfConfigFilesThisExecution);
-                // save upon regular session save at the very end
+                session.save(); // save directly to avoid parallel executions of the exact same config in AEMaaCS 
             } else {
                 LOG.debug("Node {} does not exist yet", HistoryUtils.ACHISTORY_PATH);
             }


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/impl/AcConfigChangeTracker.java
```diff
@@ -41,7 +41,7 @@ public boolean configIsUnchangedComparedToLastExecution(Map<String, String> conf
                 }
                 
                 acHistoryNode.setProperty(executionKey, hashOfConfigFilesThisExecution);
-                // save upon regular session save at the very end
+                session.save(); // save directly to avoid parallel executions of the exact same config in AEMaaCS
```
kwin, line 44: Although this will reduce the probability of the race condition this won't prevent it. It is still possible that two pods almost immediately execute the same config. I think we need rather a repository marker at the beginning of the execution indicating that some pod is starting with processing the config.... The other pod then does not start processing if it sees that marker node/property.

ghenzler: yes, this just improves the situation (but by a fair bit). 

What I would like to avoid is the block parallel executions entirely, so the additional new marker (the current marker cannot be used as it relies already the config file content hash) would have to be set in the beginning and removed after the above marker is set (and for intermediate exception cases, it has be removed in a finally block somewhere).

PR #619, opened 2021-11-16
#618 Allow startup hook to run asynchronously for mutable content

### LINKED ISSUE #618: Allow the startup hook to run asynchronously for mutable content
The AC Tool potentially can run a fairly long time (>1 minute) for mutable content if there is a large magnitude. 

For those cases and when being on AEMaaCS having the AC Tool applied to mutable content via Startup Hook, there should be an option to run the AC Tool asynchronously (to not block startup of pods during AEMaaCS deployment)

To start with this option should not be the default, possibly it makes sense to make it the default later.

ghenzler:
Closes #618

### DIFF
--- a/accesscontroltool-startuphook-bundle/src/main/java/biz/netcentric/cq/tools/actool/startuphook/impl/AcToolStartupHookServiceImpl.java
+++ b/accesscontroltool-startuphook-bundle/src/main/java/biz/netcentric/cq/tools/actool/startuphook/impl/AcToolStartupHookServiceImpl.java
@@ -41,6 +41,8 @@ import biz.netcentric.cq.tools.actool.history.impl.HistoryUtils;
 public class AcToolStartupHookServiceImpl {
     private static final Logger LOG = LoggerFactory.getLogger(AcToolStartupHookServiceImpl.class);
 
+    private static final String THREAD_NAME_ASYNC = "actool-async";
+
     @ObjectClassDefinition(name = "AC Tool Startup Hook", description = "Applies AC Tool config automatically upon startup (depending on configuration/runtime)")
     public static @interface Config {
         public enum StartupHookActivation {
@@ -49,8 +51,11 @@ public class AcToolStartupHookServiceImpl {
 
         @AttributeDefinition(name = "Activation Mode", description = "Apply on startup - CLOUD_ONLY autodetects the cloud (by missing OSGi installer bundle) and only runs on startup if deployed in the cloud. ALWAYS can be useful for local testing. NEVER disables AC Tool runs on startup entirely.")
         StartupHookActivation activationMode() default StartupHookActivation.CLOUD_ONLY;
+
+        @AttributeDefinition(name = "Async for Mutable Content", description = "Will execute on the mutable content asynchronously using a Sling Job")
+        boolean runAsyncForMutableConent() default false;
     }
-    
+
     @Reference(policyOption = ReferencePolicyOption.GREEDY)
     private AcInstallationService acInstallationService;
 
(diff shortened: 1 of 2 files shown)


accesscontroltool-startuphook-bundle/src/main/java/biz/netcentric/cq/tools/actool/startuphook/impl/AcToolStartupHookServiceImpl.java
```diff
@@ -49,13 +58,21 @@
 
         @AttributeDefinition(name = "Activation Mode", description = "Apply on startup - CLOUD_ONLY autodetects the cloud (by missing OSGi installer bundle) and only runs on startup if deployed in the cloud. ALWAYS can be useful for local testing. NEVER disables AC Tool runs on startup entirely.")
         StartupHookActivation activationMode() default StartupHookActivation.CLOUD_ONLY;
+
+        @AttributeDefinition(name = "Async for Mutable Content", description = "Will execute on the mutable content asynchronously using a Sling Job")
+        boolean runAsyncForMutableConent() default false;
```
ghenzler, line 56: Default `false` to start with

accesscontroltool-startuphook-bundle/src/main/java/biz/netcentric/cq/tools/actool/startuphook/impl/AcToolStartupHookServiceImpl.java
```diff
@@ -64,36 +81,76 @@ public void activate(BundleContext bundleContext, Config config) {
...
-                        + (relevantPathsForInstallation.isEmpty() ? "all paths" : "paths " + relevantPathsForInstallation) + "...");
-                acInstallationService.apply(null, relevantPathsForInstallation.toArray(new String[relevantPathsForInstallation.size()]),
-                        true);
-                LOG.info("AC Tool Startup Hook done. (start level " + RuntimeHelper.getCurrentStartLevel(bundleContext) + ")");
 
-                copyAcHistoryToOrFromApps(isCloudReady);
-
-            } catch (RepositoryException e) {
-                LOG.error("Exception while triggering AC Tool on startup: " + e, e);
+            if (runAsyncForMutableConent && isCompositeNodeStore) {
```
kwin, line 89: I would allow async execution if not on composite store. The general option for activationMode is sufficient here as well.

accesscontroltool-startuphook-bundle/src/main/java/biz/netcentric/cq/tools/actool/startuphook/impl/AcToolStartupHookServiceImpl.java
```diff
@@ -64,36 +81,76 @@ public void activate(BundleContext bundleContext, Config config) {
...
+
+        } catch (RepositoryException e) {
+            LOG.error("Exception while triggering AC Tool on startup: " + e, e);
+        }
+    }
+
+    private void runAcToolAsync(final BundleContext bundleContext, final boolean isCloudReady) {
+
+        final AcToolStartupHookServiceImpl startupHook = this;
+        JobConsumer acToolJobConsumer = new JobConsumer() {
```
kwin: I would not go for sling jobs but rather for lightweight threads or sling scheduler

kwin (changes requested):

accesscontroltool-startuphook-bundle/src/main/java/biz/netcentric/cq/tools/actool/startuphook/impl/AcToolStartupHookServiceImpl.java
ghenzler: @kwin for the image build we cannot rely on async - `isCompositeNodeStore` is false during image build

kwin: Right, let us a add a comment here and some output in the log to state this explicitly

accesscontroltool-startuphook-bundle/src/main/java/biz/netcentric/cq/tools/actool/startuphook/impl/AcToolStartupHookServiceImpl.java
ghenzler: IIRC the running job will prevent shutdown of the instance while the job is running while the scheduler/a thread might not?

kwin: No, even Sling Jobs will not prevent shutdown, but the persistence make them start again the next time an instance connected to the same repo is started (https://sling.apache.org/documentation/bundles/apache-sling-eventing-and-job-handling.html#jobs-guarantee-of-processing).

ghenzler: fair enough... to keep it simple I'd just go for a thread then (the scheduler is not available during image build, so we would have to decouple via an optional service, not worth the effort, a simple thread should do the job)

accesscontroltool-startuphook-bundle/src/main/java/biz/netcentric/cq/tools/actool/startuphook/impl/AcToolStartupHookServiceImpl.java
ghenzler: added a log message in f6f81249d0cf8142e9c87dc82cfed6e637ca5320

accesscontroltool-startuphook-bundle/src/main/java/biz/netcentric/cq/tools/actool/startuphook/impl/AcToolStartupHookServiceImpl.java
```diff
@@ -64,43 +69,65 @@ public void activate(BundleContext bundleContext, Config config) {
...
+
+    private void runAcToolAsync(final List<String> relevantPathsForInstallation, final int currentStartLevel, final boolean isCloudReady) {
+
+        final AcToolStartupHookServiceImpl startupHook = this;
+        new Thread(new Runnable() {
+            @Override
+            public void run() {
+                startupHook.runAcTool(relevantPathsForInstallation, currentStartLevel, isCloudReady);
+            }
+        }, THREAD_NAME_ASYNC).start();
```
ghenzler, line 121: simplified using a thread (f6f81249d0cf8142e9c87dc82cfed6e637ca5320)

kwin (approved):

Issue #648, opened 2022-10-11
Excessive usage of memory during saving merged configs


otarsko:
Storing merged configuration is done via serialization of AcConfiguration with Snakeyaml. With big amount of authorizables this consuming big amount of memory, what can cause OutOfMemory.

Used version of ACTool 3.0.4
![image](https://user-images.githubusercontent.com/3098837/195028371-0cec8f74-3ea5-4de1-b8c6-a6618cab9e3f.png)

```
"[0:0:0:0:0:0:0:1] [1665467894701] POST /mnt/overlay/netcentric/actool/content/overview/content/items/actoolpanel HTTP/1.1-ACTool-Config-Worker" #108 prio=5 os_prio=31 cpu=274178.06ms elapsed=419.03s tid=0x00007faadec6c800 nid=0x28207 runnable  [0x00007000170c2000]
   java.lang.Thread.State: RUNNABLE
        at java.util.HashMap.getNode(java.base@11.0.7/HashMap.java:576)
        at java.util.HashMap.containsKey(java.base@11.0.7/HashMap.java:591)
        at java.util.HashSet.contains(java.base@11.0.7/HashSet.java:204)
        at org.yaml.snakeyaml.serializer.Serializer.serializeNode(Serializer.java:160)
        at org.yaml.snakeyaml.serializer.Serializer.serializeNode(Serializer.java:206)
        at org.yaml.snakeyaml.serializer.Serializer.serializeNode(Serializer.java:206)
        at org.yaml.snakeyaml.serializer.Serializer.serialize(Serializer.java:113)
        at org.yaml.snakeyaml.Yaml.dumpAll(Yaml.java:270)
        at org.yaml.snakeyaml.Yaml.dumpAll(Yaml.java:237)
        at org.yaml.snakeyaml.Yaml.dump(Yaml.java:215)
        at biz.netcentric.cq.tools.actool.configreader.YamlConfigurationMerger.getMergedConfigurations(YamlConfigurationMerger.java:228)
        at biz.netcentric.cq.tools.actool.impl.AcInstallationServiceImpl.installConfigurationFiles(AcInstallationServiceImpl.java:286)
```

PR #649, opened 2022-10-12
#648: switch to the regular toString for logging to prevent excessive memory usage

otarsko:
Replace serialization of config from Yaml to regular toString. 
### Before (OutOfMemory ons serialization)
Size 685,768,736 B

### After (no OutOfMemory)
Size 1,572,864,032 B

## Example of serialized config in the history
image

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/history/impl/AcHistoryServiceImpl.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/history/impl/AcHistoryServiceImpl.java
@@ -83,7 +83,7 @@ public class AcHistoryServiceImpl implements AcHistoryService {
 
             String mergedAndProcessedConfig = installLog.getMergedAndProcessedConfig();
             if (StringUtils.isNotBlank(mergedAndProcessedConfig)) {
-                JcrUtils.putFile(historyNode, "mergedConfig.yaml", "text/yaml",
+                JcrUtils.putFile(historyNode, "mergedConfig.txt", "text/plain",
                         new ByteArrayInputStream(mergedAndProcessedConfig.getBytes()));
             }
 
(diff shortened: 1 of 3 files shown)


kwin (approved):
LGTM, thanks. Is the dump from above with or without #647?

otarsko:
@kwin that's without - testing of #647 is pending (and will be done on top of these changes)

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/history/impl/AcHistoryServiceImpl.java
```diff
@@ -83,7 +83,7 @@ public void persistHistory(PersistableInstallationLogger installLog) {
 
             String mergedAndProcessedConfig = installLog.getMergedAndProcessedConfig();
             if (StringUtils.isNotBlank(mergedAndProcessedConfig)) {
-                JcrUtils.putFile(historyNode, "mergedConfig.yaml", "text/yaml",
+                JcrUtils.putFile(historyNode, "mergedConfig.txt", "text/plain",
```
ghenzler, line 86: so this is a shame to have only a "toString" representation saved - isn't it possible to somehow generate the yaml without too much memory impact? (IIRC today snippets of this file can be used to test things and do some "round-tripping")

kwin: suggestions welcome, SnakeYAML itself has huge overhead!

ghenzler: I suppose we could fairly easily write a toString method, that creates valid yaml (instead of just using `ToStringBuilder`) - but we can also do that separately, not the most important to have immediately

ghenzler (approved):

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/history/impl/AcHistoryServiceImpl.java
otarsko: @ghenzler I've created #651 to improve toString. Hope it's fine.

PR #658, opened 2023-02-17
Make FileVault >= 3.5.4 behave as lenient as earlier versions

### LINKED ISSUE #657: Invalid initial content leads to no groups and ACEs being generated after the latest AEMaaCS release 2023.1.10912
After AEMaaCS release, AC tool doesn't generate all groups from yaml files and final dump doesn't match yaml files from codebase. There not any errors in logs.

kwin:
This closes #657

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/ContentHelper.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/ContentHelper.java
@@ -66,7 +66,7 @@ public class ContentHelper {
 
             try {
 
-                String parentPath = StringUtils.substringBeforeLast(path, "/");
+                String parentPath = Text.getRelativeParent(path, 1);
                 if (!session.nodeExists(parentPath)) {
                     history.incMissingParentPathsForInitialContent();
                     history.addVerboseMessage(LOG, "Parent path " + parentPath + " missing for initial content at " + path);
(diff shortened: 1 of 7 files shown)


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/ContentHelper.java
```diff
@@ -137,7 +136,7 @@ public static void importContent(final Session session, final String path,
 
             ImportOptions importOptions = new ImportOptions();
             importOptions.setAutoSaveThreshold(Integer.MAX_VALUE); // IMPORTANT: this disables saving the session in Importer
-            importOptions.setStrict(true);
+            importOptions.setStrict(false);
```
ghenzler: @kwin So this is the real change, that makes the difference? (apart from the test and test infra changes). Not using strict is ok for this use case I think

ghenzler (approved):

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/ContentHelper.java
```diff
@@ -77,8 +77,7 @@ public static boolean createInitialContent(final Session session, final Installa
                 history.addMessage(LOG, "Created initial content for path " + path);
                 return true;
             } catch (Exception e) {
```
kwin, line 79: distinguish between PathNotFoundException and everything else. The first one should emit a WARN only.

Issue #681, opened 2023-12-20
/oak:index/repACL-custom-1 triggers an AEMaaCS Code Validation error - Standard format not followed 


diogopedreira:
I've noticed that, at least from Cloud Manager's December release, the Code Scanning that runs as part of any pipeline, started complaining with the fact that the /oak:index/repACL-custom-1 is not following the standards (finding entry below)

> 2023-12-20T11:07:52.15Z,**Netcentric:accesscontroltool-oakindex-package:3.0.10**,0,"**Index definition detected at /oak:index/repACL-custom-1**. **Fully custom index names need to follow the pattern <prefix>.<name>-custom-<version>**, e.g. /oak:index/acme.indexName-custom-1, to be compliant with AEM as a Cloud Service",Code Smell,Minor,,IndexValidFullyCustomName,"aem,cloud-service-compatibility",**https://www.adobe.com/go/aem_cmcq_indexvalidfullycusto_en**

Reworking the index name, in a similar action like the one made in https://github.com/Netcentric/accesscontroltool/pull/550 (targeting just the index itself) and https://github.com/Netcentric/accesscontroltool/pull/638, would fix the issue.

@kwin please let me know if this would be enough and I would gladly help with the change if you want me to.

kwin:
Seems like the guidelines have changed again. Now https://experienceleague.adobe.com/en/docs/experience-manager-cloud-service/content/operations/indexing#index-names states:

> Fully custom index: It is possible to create an entirely new index from scratch. Their name must have a prefix to avoid naming conflicts. For instance: /oak:index/acme.product-1-custom-2, where the prefix is acme.

kwin:
@diogopedreira Can you come up with a PR for refactoring the name again?

diogopedreira:
@kwin, done https://github.com/Netcentric/accesscontroltool/pull/682. I've made a comment assuming a version number that I'm not 100% sure, as such let me know if it's ok or not so I can adjust and please poke me there if you want me to address any other topic.

kwin:
I cannot reproduce the issue with 3.0.10 on Cloud Manager with AEM.
I embedded `accesscontroltool-oakindex-package` and it is installed fine, and I see no related issue in the code check report. I assume this was a false positive from Adobe and meanwhile fixed in Cloud Manager. Please reopen if you still see an issue @diogopedreira

PR #682, opened 2023-12-20
#681 Updated index name to match Cloud Manager new standards

diogopedreira:
Following https://github.com/Netcentric/accesscontroltool/issues/681, the existing index has been renamed in order to be compliant with Adobe's guidelines - https://docs.mktossl.com/docs/experience-manager-cloud-service/content/operations/indexing.html?lang=en#index-names.

Changes:

- Updated index name
- Adjusted filter.xml to point to the new name and updated the existing rules to map the approach applied in https://github.com/Netcentric/accesscontroltool/pull/638/files
- Updated QueryHelper.java to refer to the new index name.


kwin:
According to https://docs.mktossl.com/docs/experience-manager-cloud-service/content/operations/indexing.html?lang=en#index-names the pattern needs to be

`<prefix>.<indexName>-<productVersion>-custom-<customVersion>`

diogopedreira:
> According to https://docs.mktossl.com/docs/experience-manager-cloud-service/content/operations/indexing.html?lang=en#index-names the pattern needs to be
> 
> `<prefix>.<indexName>-<productVersion>-custom-<customVersion>`

@kwin I've followed the pattern that was on the error log and it seems that there are some discrepancies on the standard, on the build result I get

2023-12-22T13:55:26.399Z,Netcentric:accesscontroltool-oakindex-package:3.0.10,0,"Index definition detected at /oak:index/repACL-custom-1. Fully custom index names need to follow the pattern **<prefix>.<name>-custom-<version>, e.g. /oak:index/acme.indexName-custom-1,** to be compliant with AEM as a Cloud Service",Code Smell,Minor,,IndexValidFullyCustomName,"aem,cloud-service-compatibility",https://www.adobe.com/go/aem_cmcq_indexvalidfullycusto_en 

The URL that they present is now returning a 404 and in here we have a different definition:
https://experienceleague.adobe.com/docs/experience-manager-cloud-service/content/operations/indexing.html?lang=en#index-management-without-rolling-deployments

In any case I will take the one you mentioned as it seems to be the new one.

kwin:
The pattern without product version makes a lot more sense to me. Thanks for pointing out the discrepancies in the documentation. If this version does not emit any warning in Cloud Manager I am fine with that

kwin:
@diogopedreira Have you tried to deploy with the node name as proposed in this PR? Does it emit any warnings in CloudManager?

kwin:
The unmodified index definition does no longer emit any issue in the code checks in Cloud Manager, therefore no need to change it right now.

PR #692, opened 2024-03-28
#690 - Grant access to StringEscapeUtils.escapeXml11

### LINKED ISSUE #690: When using variables in initialContent, allow to properly escape them
https://github.com/Netcentric/accesscontroltool/blob/develop/docs/AdvancedFeatures.md#providing-initial-content provides the ability to create content on the fly 

For cases where the whole string is given as literal, it just needs to be valid xml. For cases that EL expressions `${..}` are used, there is at the moment no way to encode it correctly without using a long chain of `replace()` functions. Therefore https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringEscapeUtils.html#escapeXml11-java.lang.String- shall be made available and the documentation be updated to encourage the use of this function for the `intitialContent` case.

bpeptena:
This commit fixes #690

### DIFF
--- a/docs/AdvancedFeatures.md
+++ b/docs/AdvancedFeatures.md
@@ -24,6 +24,7 @@ Function Signature | Description
 `startsWith(String str, String prefix)`| [`StringUtils.startsWith(...)`](https://commons.apache.org/proper/commons-lang/javadocs/api-3.3/org/apache/commons/lang3/StringUtils.html#startsWith(java.lang.CharSequence,%20java.lang.CharSequence))
 `length(String string)`| [`StringUtils.length(...)`](https://commons.apache.org/proper/commons-lang/javadocs/api-3.3/org/apache/commons/lang3/StringUtils.html#length(java.lang.CharSequence))
 `defaultIfEmpty(String str, String default)` | [`StringUtils.defaultIfEmpty(...)`](https://commons.apache.org/proper/commons-lang/javadocs/api-3.3/org/apache/commons/lang3/StringUtils.html#defaultIfEmpty(T,%20T))
+`escapeXml11(String str)` | [`StringEscapeUtils.escapeXml11(...)`](https://commons.apache.org/proper/commons-lang/javadocs/api-3.3/org/apache/commons/lang3/StringEscapeUtils.html#escapeXml11(java.lang.String))
 `containsItem(List<String> list, String item)`| Returns `true` if the item is contained in the given list.
 `containsAnyItem(List<String> list, List<String> items)`| Returns `true` if any of the items is contained in the given list.
 `containsAllItems(List<String> list, List<String> items)`| Returns `true` if all of the items are contained in the given list (independent of their order).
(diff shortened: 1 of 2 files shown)


kwin:
@bpeptena Looks good, can you add a unit test for it?

Positronic-Brain:
I'd also document it under /docs/AdvancedFeatures.md

Positronic-Brain:
AH, already there- Please disregard my comment - leaving my old comment as proof of my near-blindness >_<

docs/AdvancedFeatures.md
```diff
@@ -24,6 +24,7 @@ Function Signature | Description
 `startsWith(String str, String prefix)`| [`StringUtils.startsWith(...)`](https://commons.apache.org/proper/commons-lang/javadocs/api-3.3/org/apache/commons/lang3/StringUtils.html#startsWith(java.lang.CharSequence,%20java.lang.CharSequence))
 `length(String string)`| [`StringUtils.length(...)`](https://commons.apache.org/proper/commons-lang/javadocs/api-3.3/org/apache/commons/lang3/StringUtils.html#length(java.lang.CharSequence))
 `defaultIfEmpty(String str, String default)` | [`StringUtils.defaultIfEmpty(...)`](https://commons.apache.org/proper/commons-lang/javadocs/api-3.3/org/apache/commons/lang3/StringUtils.html#defaultIfEmpty(T,%20T))
+`escapeXml11(String str)` | [`StringEscapeUtils.escapeXml11(...)`](https://commons.apache.org/proper/commons-lang/javadocs/api-3.3/org/apache/commons/lang3/StringEscapeUtils.html#escapeXml11(java.lang.String))
```
ghenzler, line 27: Sorry to only look at this now - @bpeptena I would prefer to expose this method as `escapeXml()` to the yaml files and I would keep the fact, that `StringEscapeUtils.escapeXml11()` is called under the hood an implementation detail (it's fine to mention it in the docu, but I would not force all users to write the method with `...11` in the yaml files. 

Especially because XML is not exactly a fast moving technology, I don't expect a `StringEscapeUtils.escapeXml12()` or even `StringEscapeUtils.escapeXml20()` to pop up soon ;-)

kwin: There is a difference between XML 1.0 and 1.1 with respect to escape characters. We need the 1.0 variant for FileVault XML. Also making it explicit in the method name exposed in EL is IMHO a good thing.

kwin:
Superseded by #712

docs/AdvancedFeatures.md
ghenzler: I'd prefer to have the AC Tool select the correct encoding, as developer I just want to encode the xml (and call the method `encodeXml()`, then the AC Tool should select the correct way (maybe even depending on the AEM version in the future, but most likely will just be `StringEscapeUtils.escapeXml10()` as Docview requires that)

kwin: How should ACTool know the context? In some contexts XML 1.1 might make sense. The EL Parser doesn't really consider where this is being used (if with `initialContent` or somewhere else and we should IMHO leave it context free.

PR #699, opened 2024-04-25
Optionally create/update managed groups in Adobe IMS

### LINKED ISSUE #698: Create groups in Adobe IMS automatically
With AEMaaCS the users are managed externally in Adobe IMS. Groups can be either assigned in AEM or IMS but the latter has the advantage that adding a user and granting permissions happen in the same tool. It would be beneficial if new groups are automatically created in IMS by the ACTool. That requires implementation of this API: https://adobe-apiplatform.github.io/umapi-documentation/en/api/usergroupActionCommands.html#createUserGroup

kwin:
This leverages the UMAPI

This closes #698


kwin:
I am puzzled by this error message in the IT:
`errorCode=error.usergroup.not_found, requestID=null, index=0, step=0, message=Usergroup not found: testGroup, user=testGroup`

It is expected that the user group does not exist yet, if I want to create it: https://adobe-apiplatform.github.io/umapi-documentation/en/api/usergroupActionCommands.html#createUserGroup.

Update: Raised a support ticket for this in https://adminconsole.adobe.com/FA907D44536A3C2B0A490D4D@AdobeOrg/support/support-cases/E-001223901.

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthorizableInstallerServiceImpl.java
```diff
@@ -65,6 +66,7 @@
 import biz.netcentric.cq.tools.actool.configmodel.pkcs.Key;
 import biz.netcentric.cq.tools.actool.configmodel.pkcs.RandomPassword;
 import biz.netcentric.cq.tools.actool.crypto.DecryptionService;
+import biz.netcentric.cq.tools.actool.externalusermanagement.ExternalGroupManagement;
```
ghenzler, line 69: the code in general is a bit inconsistent regarding the naming user vs group management. 

To the AC tool is 99% used in AEM (while pure sling is also supported) - we could call it just ImsGroupManagement for the 99% or we stay generic for the 1%. But I'm fine to stay generic for the 1%, but `user` is confusing because it is really only touching groups in IMS. So I would go for `ExternalGroupManagement` consistently.

docs/Configuration.md
```diff
@@ -55,11 +55,12 @@ description | Description of the group | optional
...
-migrateFrom | a group name assigned member users are taken over from | optional
+migrateFrom | A group name assigned member users are taken over from. This is not supported for external user management. | optional
 virtual | If true, the group is *not* created in repository but only its permissions and isMemberOf relationships are made effective to other groups using the virtual group in isMemberOf. Flattens the effective group tree of users at runtime, but enlarges ACLs as set on path (instead of the virtual group, all referencing groups are listed in a particular path's ACL used by the virtual group). Useful for functionality fragments. False by default. Since v2.1.0 | optional
 unmanaged* Properties | Only use sparsely and with care, see [Advanced Features](AdvancedFeatures.md)   | optional
+syncWithExternalUserManagement | Boolean flag determining whether the group should also be created or updated in a configured external user management system. For further details refer to [Update Groups in External User Management Systems](AdvancedFeatures.md#Update Groups in External User Management Systems). Since v3.1.0. | optional
```
ghenzler: so I don't like the user management in this term  (for the reason stated in previous comment). Either `syncWithExternalGroupManagement` or maybe even better `ensureCreatedInIdP`. The latter would also allow to use this feature potentially on publish (on publish it could check for an implementation of a provider interface provided by a project to create the marked groups anywhere needed)

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthorizableInstallerServiceImpl.java
ghenzler: thinking a bit more, maybe the best package name is `idpintegration`

docs/AdvancedFeatures.md
```diff
@@ -434,6 +433,15 @@ Sometimes it is also useful to create intermediate paths that do not even contai
...
 ```
 
+## Update Groups in External User Management Systems
+
+Since ACTool 3.1.0 there is support for creating/updating groups in [Adobe IMS](https://www.adobe.com/content/dam/cc/en/trust-center/ungated/whitepapers/corporate/adobe-identity-management-services-security-overview.pdf). Those are the groups which are exposed in the Adobe Admin Console and automatically used for [AEMaaCS](https://experienceleague.adobe.com/en/docs/experience-manager-cloud-service/content/security/ims-support) and also [AEM 6.5 hosted by AMS](https://experienceleague.adobe.com/en/docs/experience-manager-learn/foundation/authentication/adobe-ims-authentication-technical-video-understand).
+
+To enable that feature, just set the property `syncWithExternalUserManagement` on the groups in the YAML file to `true`.
+In addition an OSGi configuration for the leveraged [UMAPI](https://adobe-apiplatform.github.io/umapi-documentation/en/) needs to be provided in the configuration PID `biz.netcentric.cq.tools.actool.externalusermanagement.ims.IMSUserManagement`. This configuration should only be provided for run mode `author` to prevent the same groups from being created/updated multiple times.
```
ghenzler: I would be nice to leave the door open for publish

ghenzler (changes requested):

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/authorizableinstaller/impl/AuthorizableInstallerServiceImpl.java
kwin: user management for me comprises both user and group management (also the API from Adobe is called User Management API, UMAPI). IDP for me is only the authentication part but not the management part (compare with https://www.adobe.com/content/dam/cc/en/trust-center/ungated/whitepapers/corporate/adobe-identity-management-services-security-overview.pdf)

kwin: The idea was to add user management class in the same package (in order to export users if there is the need in the future).

docs/Configuration.md
kwin: see my comments before, the idea was to have a flag which could be used both in groups and users (although right now is only supported in groups). For me the most important aspect of this flag is `external`.

kwin: https://www.adobe.com/content/dam/cc/en/trust-center/ungated/whitepapers/corporate/adobe-identity-management-services-security-overview.pdf

kwin: The idea is to use the same flag to potentially sync with e.g. an LDAP or another user directory.

ghenzler: At the moment we are missing a full example, currently a group definition would look as follows:

```
    - example-group:
        - name: "Example Group"
          externalId: "example-group:ims"
          path: "/home/groups/example"
```
maybe the most straight forward, simple and concise way of configuring it is the following:

```
    - example-group:
        - name: "Example Group"
          externalId: "example-group:ims"
          externalSync: true
          path: "/home/groups/example"
```
having it prefixed with `external` groups it visually nicely with existing `externalId`

kwin: I don't think a full example is necessary with the description of both fields `externalSync` and `externalId` now, right? This is anyhow not the place where to put such an example...

docs/AdvancedFeatures.md
kwin: I do :-)

ghenzler (approved):

PR #704, opened 2024-05-02
Add functionality to export users in CSV format

### LINKED ISSUE #443: UI confusing in http://localhost:4502/mnt/overlay/netcentric/actool/content/overview.html/
The newly added Touch UI is confusing with respect to the order of buttons. 
Screenshot 2020-03-27 at 11 28 15

The button "Download Dump" is on the same line as "Configuration Root Path" although it has nothing to do with it.

I would propose to place the "Download Dump" in a dedicated row, so that it is clear that this action cannot be parameterized! Also it would be good to explain whether the dump is by authorizable/path (and what that means).

The same applies to the WebConsole at http://localhost:4502/system/console/actool

### LINKED ISSUE #703: Export Users in CSV format (compatible with Admin Console)
When you migrate from AEM Classic to Cloud you often want to leverage Adobe IMS for maintaining users and their direct group memberships. To ease this migration the JMX Console as well as the Web UI (/mnt/overlay/netcentric/actool/content/overview.html/actool) should provide a way to create a compliant CSV file (https://helpx.adobe.com/enterprise/using/bulk-upload-users.html#csv-format) which can afterwards be imported in the Admin Console.

kwin:
This format is compatible with
https://helpx.adobe.com/enterprise/using/bulk-upload-users.html#csv-format

Fix Web Console deployment with sling-m-p

This closes #703
This closes #443

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/AcToolUiService.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/AcToolUiService.java
@@ -4,18 +4,33 @@ import static org.apache.commons.lang3.StringEscapeUtils.escapeHtml4;
 
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.io.UncheckedIOException;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.text.SimpleDateFormat;
 import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
 
+import javax.jcr.RepositoryException;
+import javax.jcr.Value;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.felix.webconsole.WebConsoleConstants;
+import org.apache.jackrabbit.api.security.user.Group;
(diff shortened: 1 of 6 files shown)


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/AcToolUiService.java
```diff
@@ -44,11 +61,15 @@ public class AcToolUiService {
...
 
-    static final String PATH_SEGMENT_DUMP = "dump.yaml";
+    static final String SUFFIX_DUMP_YAML = "dump.yaml";
+    static final String SUFFIX_USERS_CSV = "users.csv";
 
     @Reference(policyOption = ReferencePolicyOption.GREEDY)
     private ConfigDumpService dumpService;
 
+    @Reference(policyOption = ReferencePolicyOption.GREEDY)
+    private UserProcessor userExporter;
```
ghenzler: What about calling the interface also`UserExpoerter`? Also would call the package of that service `userexporter`

kwin: This service does not export (in the sense of serializing or converting) something. The actual export is done within this service.

kwin: Just rename variable to userProcessor.

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/ui/AcToolUiService.java
```diff
@@ -93,18 +141,60 @@ protected void doPost(final HttpServletRequest req, final HttpServletResponse re
         }
     }
 
+    
+    private boolean mayApplyConfig(HttpServletRequest req) {
```
kwin: hasAccessToFelixWebConsole()

PR #720, opened 2024-06-04
Improve detection of index for "rep:ACL" nodes

### LINKED ISSUE #714: Ensure the index for rep:ACL is correctly detected
In AEMaaCS there is currently the issue that even though the oak index `/oak:index/repACL-custom-1` is installed, QueryHelper [1] takes the wrong query for ACEs instead of ACLs, and for our case we are above the 100,000 query limit with that. [2] should really give access to  `/oak:index/repACL-custom-1`, but `session.nodeExists(OAK_INDEX_PATH_REP_ACL)` returns false.

[1] https://github.com/Netcentric/accesscontroltool/blob/653d2acba53b5087edf703a25bdcd20f43fd3edf/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/QueryHelper.java#L101-L106
[2] https://github.com/Netcentric/accesscontroltool/blob/653d2acba53b5087edf703a25bdcd20f43fd3edf/accesscontroltool-package/src/main/jcr_root-cloud/apps/netcentric/actool/config/org.apache.sling.jcr.repoinit.RepositoryInitializer~actool.config#L3

kwin:
Rely on EXPLAIN MEASURE and evaluate the estimated cost.

This closes #714

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/QueryHelper.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/QueryHelper.java
@@ -10,6 +10,7 @@ package biz.netcentric.cq.tools.actool.helper;
 
 import static biz.netcentric.cq.tools.actool.history.impl.PersistableInstallationLogger.msHumanReadable;
 
+import java.io.IOException;
 import java.util.Collection;
 import java.util.HashSet;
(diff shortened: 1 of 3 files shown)


kwin:
@ghenzler Maybe you can take it up from there, the IT still is incorrect because by default it uses the NodeTypeIndex with cost = 5.0 (which does not exist in AEMaaCS).

PR #731, opened 2024-06-13
Also interpret cost JSON object in case it is returned by oak search engine

### LINKED ISSUE #730: Invalid index detection in the cloud
For the automatic index detection from #714, it turns out that locally the cost is returned as 
`cost: { "s": 189540.0 }`
but in the cloud as 
` cost: { "s": { perEntry: 1.0, perExecution: 1.0, count: 47814 } }`
this seems to stem from https://github.com/apache/jackrabbit-oak/blob/5d10c51ff53dc05e2d36b2491689412ccd17b15d/oak-core/src/main/java/org/apache/jackrabbit/oak/query/ast/SelectorImpl.java#L448C34-L448C52

I'm not sure why the index is only a `AdvancedQueryIndex` in the cloud, to be checked.

The reason why the impl of #714 does not throw an exception is that https://www.javadoc.io/doc/com.fasterxml.jackson.core/jackson-databind/2.8.5/com/fasterxml/jackson/databind/JsonNode.html#asDouble(double) never throws an exception but always just returns the default if  a double cannot be coerced (even for a map instead of a skalar)

ghenzler:
Closes #730

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/QueryHelper.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/QueryHelper.java
@@ -48,11 +48,15 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 public class QueryHelper {
+
     public static final Logger LOG = LoggerFactory.getLogger(QueryHelper.class);
 
     private static final String ROOT_REP_POLICY_NODE = "/rep:policy";
     private static final String ROOT_REPO_POLICY_NODE = "/" + Constants.REPO_POLICY_NODE;
     private static final String HOME_REP_POLICY = "/home/rep:policy";
+
+    private static final String EXPLAIN_QUERY_FOR_ACL_INDEX = "EXPLAIN MEASURE SELECT * FROM [rep:ACL] AS s WHERE ISDESCENDANTNODE([/])";
+    private static final String KEY_COST_PER_EXECUTION = "perExecution";
 
     /** every query cost below that threshold means a dedicated index exists, above that threshold means: fallback to traversal */
     private static final double COST_THRESHOLD_FOR_QUERY_INDEX = 100d;
@@ -110,8 +114,8 @@ public class QueryHelper {
             boolean indexForRepACLExists = false;
             try {
                 indexForRepACLExists = hasQueryIndexForACLs(session);
-            } catch(IOException|RepositoryException e) {
-                LOG.warn("Cannot figure out if query index for rep:ACL nodes exist", e);
+            } catch(Exception e) {
+                LOG.warn("Could not detect if query index for rep:ACL nodes exists: {}", e.getMessage(), e);
             }
             LOG.debug("Index for repACL exists: {}",indexForRepACLExists);
             String queryForAClNodes = indexForRepACLExists ? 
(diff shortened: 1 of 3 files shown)


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/QueryHelper.java
```diff
@@ -143,14 +145,9 @@ static boolean hasQueryIndexForACLs(final Session session) throws RepositoryExce
         Row row = queryResult.getRows().nextRow();
         // inspired by https://github.com/apache/jackrabbit-oak/blob/cc8adb42d89bc4625138a62ab074e7794a4d39ab/oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/query/QueryTest.java#L1092
         String plan = row.getValue("plan").getString();
-        String costJson = plan.substring(plan.lastIndexOf('{'));
-        
-        // use jackson for JSON parsing
-        ObjectMapper mapper = new ObjectMapper();
+        String costJsonStr = plan.substring(plan.lastIndexOf('{'));
```
kwin: Probably we should log the JSON (on debug/verbose level) in case it changes in the future, so that we can more easily come up with a fix.

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/QueryHelper.java
```diff
@@ -160,6 +157,18 @@ static boolean hasQueryIndexForACLs(final Session session) throws RepositoryExce
...
     }
 
+    static double getCostFromJsonStr(String jsonStr) throws JsonProcessingException, IOException {
+        // use jackson for JSON parsing
+        ObjectMapper mapper = new ObjectMapper().configure(JsonParser.Feature.ALLOW_UNQUOTED_FIELD_NAMES, true);
+
+        // read the json strings and convert it into JsonNode
+        JsonNode node = mapper.readTree(jsonStr);
+        
+        JsonNode sNode = node.get("s");
```
kwin, line 172: We need some more null checks here and a unit test with invalid JSON.

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/QueryHelper.java
```diff
@@ -160,6 +157,18 @@ static boolean hasQueryIndexForACLs(final Session session) throws RepositoryExce
...
+        double cost = sNode.isContainerNode() ? sNode.get("perExecution").asDouble(Double.MAX_VALUE) : sNode.asDouble(Double.MAX_VALUE);
```
kwin: let us null check sNode.get("perExecution") as well.

kwin (changes requested):

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/QueryHelper.java
```diff
@@ -108,8 +114,8 @@ public static Set<String> getRepPolicyNodePaths(final Session session,
             boolean indexForRepACLExists = false;
             try {
                 indexForRepACLExists = hasQueryIndexForACLs(session);
-            } catch(IOException|RepositoryException e) {
-                LOG.warn("Cannot figure out if query index for rep:ACL nodes exist", e);
+            } catch(Exception e) {
+                LOG.warn("Could not detect if query index for rep:ACL nodes exists: {}", e.getMessage(), e);
             }
```
ghenzler, line 119: no matter what exception is thrown this warning should be shown

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/QueryHelper.java
```diff
@@ -137,20 +143,16 @@ public static Set<String> getRepPolicyNodePaths(final Session session,
...
         QueryResult queryResult = query.execute();
         Row row = queryResult.getRows().nextRow();
         // inspired by https://github.com/apache/jackrabbit-oak/blob/cc8adb42d89bc4625138a62ab074e7794a4d39ab/oak-jcr/src/test/java/org/apache/jackrabbit/oak/jcr/query/QueryTest.java#L1092
         String plan = row.getValue("plan").getString();
-        String costJson = plan.substring(plan.lastIndexOf('{'));
-        
-        // use jackson for JSON parsing
-        ObjectMapper mapper = new ObjectMapper();
+        LOG.debug("Execution plan and cost for {}: {}", EXPLAIN_QUERY_FOR_ACL_INDEX, plan);
+        String costJsonStr = StringUtils.substringAfter(plan, "cost:");
```
ghenzler, line 153: just checking for `{` is also very unsafe - conceptually better to check for the string after `cost:`.  (`plan.substring(plan.lastIndexOf('{'))` returned only parts of the cost json)

kwin (approved):
Thanks for the fixes, looks good now.

PR #735, opened 2024-06-21
Validation to ensure groups with externalId set are not used in isMemberOf

### LINKED ISSUE #732: Validate that external roles are not used to compose other roles.
There are quite some problems that can happen if  external groups (or better external roles, AEM groups that have `externalId` set) are used to build up other roles. Problems are

* If a role is open for adding member on AEM side, then roles show up in the members list
* IMS uses dynamic groups that can lead to `javax.jcr.nodetype.ConstraintViolationException: OakConstraint0077`
* In general it is good practice to clearly separate roles and fragments following https://github.com/Netcentric/accesscontroltool/blob/develop/docs/BestPractices.md#use-fragment-groups-for-functional-aspects-and-content-access

Therefore it should be validated that groups with `externalId` set are not used in `isMemberOf` clauses - if a configuration contains such a setup, an error should be thrown.

ghenzler:
Closes #732


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/configmodel/GlobalConfiguration.java
```diff
@@ -29,6 +29,7 @@ public class GlobalConfiguration {
 
     public static final String KEY_ALLOW_CREATE_OF_UNMANAGED_RELATIONSHIPS = "allowCreateOfUnmanagedRelationships";
 
+    public static final String KEY_ALLOW_EXTERNAL_GROUPS_IN_IS_MEMBER_OF = "allowExternalGroupsInIsMemberOf";
```
kwin, line 32: why so complicated? I would prefer `KEY_ALLOW_MEMBERS_IN_EXTERNAL_GROUPS`

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/validators/ExternalGroupsInIsMemberOfValidator.java
```diff
@@ -0,0 +1,77 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(ExternalGroupsInIsMemberOfValidator.class);
+
+    public void validateIsMemberOfConfig(AcConfiguration acConfiguration, PersistableInstallationLogger installLog,
+            GlobalConfiguration globalConfiguration) throws InvalidExternalGroupUsageValidationException {
+        List<String> externalGroupsValidationResults = checkIsMemberOfConfigsOfAllAuthorizables(acConfiguration);
+        if (!externalGroupsValidationResults.isEmpty()) {
+
+            externalGroupsValidationResults.stream().forEach(m -> installLog.addWarning(LOG, m));
+
+            String validationMsg = "Found " + externalGroupsValidationResults.size() + " group(s) that use external groups in isMemberOf. ";
```
kwin: This may be user's as well! Again I would emit it the other way around:
`Found x authorizables as member of external group y`. Also the message should include which external group that is and which authorizables are members.

docs/Configuration.md
```diff
@@ -52,7 +52,7 @@ property | comment | required
...
...upport#aem-configuration). This establishes a connection between an (internal) JCR group and an externally managed group (and is persisted in the group's node in the property `rep:externalId`). The value has to be in format `<external-id>;<provider-name>`.  How the external ID and provider name look like is *External Identity Provider dependent*: For **Adobe IMS** it usually is `<groupId>;ims` while for **Oak LDAP** it usually is `<LDAP-DN>;<IDP-NAME>` where LDAP-DN is the full distinguished name and IDP-NAME is configured in OSGI config PID `org.apache.jackrabbit.oak.security.authentication.ldap.impl.LdapIdentityProvider` property `provider-name`. LDAP Example: `externalId: "cn=group-name,ou=mydepart,ou=Groups,dc=comp,dc=com;IDPNAME"`. Make sure to also set the group id according to how it is extracted by the external identify provider (configurable via OSGi configuration of the external identity provider). Using groups being synced from external sources in `isMemberOf` will cause an error to avoid problems with dynamic memberships. Use `allowExternalGroupsInIsMemberOf: true` in `global_config` if you need to override this behaviour (should be used rarely). Since v1.9.3 | optional
```
kwin: I would add a link here to https://jackrabbit.apache.org/oak/docs/security/authentication/external/dynamic.html

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/validators/ExternalGroupsInIsMemberOfValidator.java
ghenzler: line 38 adds each individual occurrence to the execution log. 

The recommendation is to use `isMemberOf` (there is a warning if you `members` is used instead of `isMemberOf`) - also at this stage member values that could be moved to the other side of the relationship have been moved via https://github.com/Netcentric/accesscontroltool/blob/b60fd036ae71cee44cfb83916a9e388d18e6cccb/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/configreader/YamlConfigurationMerger.java#L212

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/configmodel/GlobalConfiguration.java
ghenzler: https://github.com/Netcentric/accesscontroltool/pull/735#discussion_r1649158217

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/validators/ExternalGroupsInIsMemberOfValidator.java
ghenzler: > This may be user's as well! 

fixed via https://github.com/Netcentric/accesscontroltool/pull/735/commits/77fed31932f214418029d95975e45b38b0be1138

kwin (approved):

Issue #751, opened 2024-07-15
Expose StringUtils.isEmpty and StringUtils.isBlank to EL


Positronic-Brain:
For convenience, we should allow [StringUtils.isBlank ](https://commons.apache.org/proper/commons-lang/javadocs/api-3.3/org/apache/commons/lang3/StringUtils.html#isBlank(java.lang.CharSequence))and [StringUtils.isEmpty](https://commons.apache.org/proper/commons-lang/javadocs/api-3.3/org/apache/commons/lang3/StringUtils.html#isEmpty(java.lang.CharSequence)) for EL evaluation - this makes expressions more readable when evaluating strings that are either null, empty and or blank.

PR #752, opened 2024-07-15
751 Expose isEmpty and isBlank for EL evaluation

Positronic-Brain:
This would implement [issue #751](https://github.com/Netcentric/accesscontroltool/issues/751)


docs/AdvancedFeatures.md
```diff
@@ -24,6 +24,8 @@ Function Signature | Description
 `endsWith(String str, String suffix)`|[`StringUtils.endsWith(...)`](https://commons.apache.org/proper/commons-lang/javadocs/api-3.3/org/apache/commons/lang3/StringUtils.html#endsWith(java.lang.CharSequence,%20java.lang.CharSequence))
 `startsWith(String str, String prefix)`| [`StringUtils.startsWith(...)`](https://commons.apache.org/proper/commons-lang/javadocs/api-3.3/org/apache/commons/lang3/StringUtils.html#startsWith(java.lang.CharSequence,%20java.lang.CharSequence))
 `length(String string)`| [`StringUtils.length(...)`](https://commons.apache.org/proper/commons-lang/javadocs/api-3.3/org/apache/commons/lang3/StringUtils.html#length(java.lang.CharSequence))
+`isBlank(String str)` | [`StringUtils.isBlank(...)`](https://commons.apache.org/proper/commons-lang/javadocs/api-3.3/org/apache/commons/lang3/StringUtils.html#isBlank(java.lang.CharSequence))
```
kwin: You should probably add a `Since` column indicating the version since when it is supported.

kwin (changes requested):
Looks good in general, but:
- requires a minor version upgrade (please adjust the pom.xml accordingly)
- requires some info on documentation since when which method is supported.

ghenzler:
there isn't much that speaks against linking these two methods, but we could also just document the `empty` operator that EL provides out of the box (see https://docs.oracle.com/cd/E17802_01/j2ee/j2ee/1.4/docs/tutorial-update6/doc/JSPIntro7.html#wp71089)

Positronic-Brain:
Yeah, empty is a bit redundant, but since I was already exposing "isBlank" I decided to add its related method for consistency.

docs/AdvancedFeatures.md
Positronic-Brain: Fixed. I also sorted methods alphabetically because it was bugging me out,

ghenzler:
@Positronic-Brain  so I just checked and `empty '   '` really does not return true (which I thought would be the case for EL). So fair enough to add the two methods, but I probably would also add https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#defaultIfBlank-T-T- then (as we have `defaultIfEmpty(..)` already)

Positronic-Brain:
> @Positronic-Brain so I just checked and `empty ' '` really does not return true (which I thought would be the case for EL). So fair enough to add the two methods, but I probably would also add https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#defaultIfBlank-T-T- then (as we have `defaultIfEmpty(..)` already)

Sounds reasonable. Will do.

kwin (approved):
LGTM now. Thanks a lot.

Issue #743, opened 2024-07-09
Improve naming of RuntimeHelper methods


kwin:
Currently the [RuntimeHelper](https://github.com/Netcentric/accesscontroltool/blob/6499e07816c37d82147b0e5954b0db15704fa400/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/runtime/RuntimeHelper.java) exposes two methods related to cloud

1. isCompositeNodeStore and
2. isCloudReadyInstance

Both namings are a bit unfortunate, because 

1. in cloud AEM is always started with a [composite node store](https://jackrabbit.apache.org/oak/docs/nodestore/compositens.html) (even during the image build phase). The difference between image build and regular server start is which mount is the default one (apps/libs or the rest) (https://github.com/apache/jackrabbit-oak/blob/8adf6aa2d931e562ec8e69bf4261a250c53fe6f3/oak-store-composite/src/main/java/org/apache/jackrabbit/oak/composite/CompositeNodeStoreService.java#L90).
2. sounds a bit like distinction between AEM Classic and AEMaaCS (SDK or Cloud), but in fact what is meant here is if the composite node store is used (where only one mount is read-write)

I propose to rename

1. `isCompositeNodeStore` to `isAppsReadOnly` 
2. `isCloudReadyInstance` to `isCompositeNodeStore`

That way it is much clearer what is meant.

kwin:
Actually I was wrong, AEM is not using the composite node store during the image build phase. The `seed` mode is just for pre-populating the default store but does not allow to write to `/apps` or `/libs`. I tried to clarify this in https://github.com/apache/jackrabbit-oak/pull/1719/files.

PR #767, opened 2024-09-09
Rename RuntimeHelper methods to match their implementations closer

bewifly:
Solves https://github.com/Netcentric/accesscontroltool/issues/743


accesscontroltool-startuphook-bundle/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
bewifly, line 1: Mockito extension for testing static methods https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mockito.html#Mocking_Final

kwin:
This is superseded by #869. @bewifly Can you have a look there?

PR #770, opened 2024-10-22
Link custom Oak restrictions in README

kwin:


kwin:
@rolandgruber-ibm FYI

PR #812, opened 2025-05-09
Fix REPLICATE_PRIVILEGE constant.

### LINKED ISSUE #811: Version 3.6.0 fails to apply the  `replicate` action
The change in #802 to remove the uber-jar has replaced the constant 
`com.day.cq.replication.Replicator.REPLICATE_PRIVILEGE` with a value of `rep:replicate`. The problem with this is that this is not a like-for-like replacement - the inbuilt constant has a value of `{http://www.day.com/crx/1.0}replicate`, which would translate to `crx:replicate` when executed. 

Changes now fail to apply replication permission correctly, as `rep:replicate` is semi-silently accepted (it does report in logs "Replicate privilege not registered" as per the catch rule), but generates no working ACE that AEM can understand.

The string value needs to be changed.

pgorny:
Fixes #811.

The removal of the uber-jar dependency in #802 has introduced an incorrect name for the privilege that is not understood by AEM.


kwin:
Thanks a lot for the PR. Will release a hotfix with this PR included next week.

PR #821, opened 2025-06-03
Support EL interpolation in group's external id

### LINKED ISSUE #819: Allow to reference the authorizable id from within the externalId property
As the group's external id in most of the cases is just the group id suffixed with something else it is tedious/error-prone to repeat the full group id in the YAML config. Instead it should be possible to refer to the group id with a dedicated variable.

This is already supported in the section `autoCreateTestUsers`: https://github.com/Netcentric/accesscontroltool/blob/develop/docs/AdvancedFeatures.md#automatically-purge-obsolete-groups-and-users

> Interpolation of group properties can be used with EL, however as $ is evaluated at an earlier stage, %{} is used here. Available is %{group.id}, %{group.name}, %{group.path} or expressions like %{split(group.path,'/')[2]}.

kwin:
Reuse logic from autoCreateTestUsers.
This closes #819


kwin:
@vitafo Please have a look

vitafo (approved):

PR #855, opened 2025-10-02
Do not pre-validate restrictions

### LINKED ISSUE #854: Metadata-driven permissions - Restrictions support
Followed the Adobe documentation on how to enable metadata driven permissions [here](https://experienceleague.adobe.com/en/docs/experience-manager-learn/assets/advanced/metadata-driven-permissions#) and while trying to define a restriction based on region metadata property (custom), I get this error:

`ERROR: Could not process yaml files / e=biz.netcentric.cq.tools.actool.validators.exceptions.InvalidRestrictionsException: Validation error while reading ACE
definition nr.2 of authorizable editor-region, this repository doesn't support following restriction(s): [region]`

Looks like these restrictions are not listed by [JackrabbitAccessControlList:getRestrictionNames](https://developer.adobe.com/experience-manager/reference-materials/6-5/javadoc/org/apache/jackrabbit/oak/spi/security/authorization/accesscontrol/AbstractAccessControlList.html#getRestrictionNames)

OSGi config to allow metadata restrictions (`com.adobe.cq.dam.assetmetadatarestrictionprovider.impl.DefaultRestrictionProviderConfiguration`):
```
{
  "restrictionPropertyNames":[
    "region"
  ],
  "restrictionContentPropertyNames":[],
  "enabled":true
}
```

ACEs:
```
- editor-region:
    - path: /content/dam/brand
      permission: deny
      actions: read
      restrictions:
        rep:ntNames: dam:Asset

    - path: /content/dam/brand
      permission: allow
      actions: read
      restrictions:
        region: europe
```

kwin:
Oak takes care of that in
https://github.com/apache/jackrabbit-oak/blob/17281282fe82d0f0c4e86d0a42ecfb20bfe404e3/oak-core/src/main/java/org/apache/jackrabbit/oak/security/authorization/accesscontrol/ACL.java#L213 as soon as you try to apply those.
Otherwise you face exceptions when using restrictions only available at run-time in cloud (but not at build time).

This closes #854


ggruianc:
@kwin there is another restrictions pre-validation in

https://github.com/Netcentric/accesscontroltool/blob/8fdc6ba31e9b0aaa4bd73c5388b1cb8668c022f6/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/validators/impl/AceBeanValidatorImpl.java#L134-L144

kwin:
> @kwin there is another restrictions pre-validation in

Thanks, that validation should be removed now in https://github.com/Netcentric/accesscontroltool/pull/855/commits/1c4b417470bd215d8dc30eaacde378614800280d

accesscontroltool-bundle/src/test/resources/testconfig.yaml
```diff
@@ -285,15 +285,12 @@
...
-         assertedException: InvalidRestrictionsException
          
        #11 wrong restriction name. correct one would be rep:glob
        - path: /content
          permission: allow
          actions: read
          restrictions:
            rep:Glob: /cq:*
-         assertedException: InvalidRestrictionsException
-
```
ghenzler, line 297: We should not just remove `assertedException` but then also the test case with the comment

ghenzler (approved):

accesscontroltool-bundle/src/test/resources/testconfig.yaml
kwin: For me this test shows that invalid restrictions do no longer lead to exceptions, therefore this is still helpful.

PR #864, opened 2026-01-14
Remove admin ACLs following Oak Best practices

Amoratinos:
As mentioned in [Oak authorization best practices](https://jackrabbit.apache.org/oak/docs/security/authorization/bestpractices.html#avoid-redundancy) avoiding duplicity by removing ACLs for administrative principals

### DIFF
--- a/accesscontroltool-apps-package/src/main/jcr_root/apps/netcentric/actool/content/overview/_rep_policy.xml
+++ b/accesscontroltool-apps-package/src/main/jcr_root/apps/netcentric/actool/content/overview/_rep_policy.xml
@@ -19,12 +19,4 @@
         jcr:primaryType="rep:DenyACE"
         rep:principalName="everyone"
         rep:privileges="{Name}[jcr:read]"/>
-    <allow1
-        jcr:primaryType="rep:GrantACE"
-        rep:principalName="admin"
-        rep:privileges="{Name}[jcr:read]"/>
-    <allow2
-        jcr:primaryType="rep:GrantACE"
-        rep:principalName="administrators"
-        rep:privileges="{Name}[jcr:read]"/>
 </jcr:root>


accesscontroltool-apps-package/src/main/jcr_root/apps/netcentric/actool/content/overview/_rep_policy.xml
```diff
@@ -19,12 +19,4 @@
...
         rep:principalName="everyone"
         rep:privileges="{Name}[jcr:read]"/>
-    <allow1
-        jcr:primaryType="rep:GrantACE"
-        rep:principalName="admin"
-        rep:privileges="{Name}[jcr:read]"/>
-    <allow2
-        jcr:primaryType="rep:GrantACE"
-        rep:principalName="administrators"
-        rep:privileges="{Name}[jcr:read]"/>
```
ghenzler, line 29: So "admin" could be removed (as the ACLs are not checked for `admin` anyway IIRC) but the group `administrators` should stay to ensure administrators can reliably access this console regardless from ACLs further up the node hierarchy? How exactly do you think this violates https://jackrabbit.apache.org/oak/docs/security/authorization/bestpractices.html#avoid-redundancy @Amoratinos ?

kwin: I think `administrators` are always configured in configuration PID `org.apache.jackrabbit.oak.security.authorization.AuthorizationConfigurationImpl` as `administrativePrincipals` next to `admin` (https://github.com/apache/jackrabbit-oak/blob/c9209db5562dddd02e56aa860cbf4f8d9a5ad955/oak-core/src/main/java/org/apache/jackrabbit/oak/security/authorization/AuthorizationConfigurationImpl.java#L118) therefore both should behave the same.

Amoratinos: As @kwin said by default admin & administrators principals are configured as administrativePrincipals. Permissions evaluation is skipped for those so could be removed

anchela: that's it! having that ACE will depending on the value of the importhaviour config option log a warning or fail the setup in AEM ootb.

ghenzler: Ok, I wasn't aware that `administrators` is configured via 
org.apache.jackrabbit.oak.security.authorization.AuthorizationConfigurationImpl to skip permission evaluation as well. I also just checked a local AEM 6.5 LTS instance (that we also need to support), also there it seem to be configured. So if everybody is confident that it will not make a difference we can go ahead with this.

PR #869, opened 2026-03-12
Use run modes to determine server type

kwin:
Deprecate other/more brittle ways

### DIFF
--- a/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/aceinstaller/BaseAceBeanInstaller.java
+++ b/accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/aceinstaller/BaseAceBeanInstaller.java
@@ -23,6 +23,7 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.function.Predicate;
 
 import javax.jcr.RepositoryException;
 import javax.jcr.Session;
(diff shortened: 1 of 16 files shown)


accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/aceinstaller/BaseAceBeanInstaller.java
```diff
@@ -107,26 +116,62 @@ public void installPathBasedACEs(
...
-                    pathsToKeep.add(path);
-                }
-            }
-            history.addMessage(LOG, "Ignoring " + readOnlyPaths.size() + " ACLs in /apps and /libs because they are ready-only (Composite NodeStore)");
+    private Set<String> filterForRelevantPaths(Set<String> paths, InstallationLogger history, Session session) {
+        if (serverType == ServerType.AEM_CLOUD_RUN) {
+            Set<String> pathsToKeep = removePathsWithPrefixes(paths, IMMUTABLE_PATH_PREFIXES);
+            history.addMessage(LOG, "Ignoring " + (paths.size() - pathsToKeep.size()) + " ACLs in " + String.join(", ", IMMUTABLE_PATH_PREFIXES) + " because they are ready-only (Composite NodeStore)");
+            return pathsToKeep;
+        } else if (serverType == ServerType.AEM_CLOUD_IMAGE_BUILD) {
```
kwin, line 124: @ghenzler Should we also skip group installation during image build?

ghenzler: we should be able to - but has to be done elsewhere and tested (this piece of code only looks at the paths)

accesscontroltool-bundle/src/main/java/biz/netcentric/cq/tools/actool/helper/runtime/RuntimeHelper.java
```diff
@@ -25,12 +28,19 @@
...
     public static final Logger LOG = LoggerFactory.getLogger(RuntimeHelper.class);
 
     private static final String INSTALLER_CORE_BUNDLE_SYMBOLIC_ID = "org.apache.sling.installer.core";
 
+    /**
+     * 
+     * @param session
+     * @return
+     * @deprecated Use {@link #getServerType(Collection)} instead
+     */
```
kwin, line 43: As not exported and not part of api, remove method instead of deprecating it.

PR #904, opened 2026-08-03
Prevent spooling of arbitrary classes via AcToolUiService

kwin:
Improve copying of content to output stream


kwin:
Not gonna fix as this affects all Felix WebConsole plugins (https://issues.apache.org/jira/browse/FELIX-6854).

